PR #1, opened 2020-12-07
Pod security standards

JimBugwadia:
initial set of policies for pod security standards


chipzoller:
> @JimBugwadia with the adding of wildcards to `pattern` objects, does it support `?*` or only `?` or `*`? The reason I thought of specifying `?*` was that the annotation would never be valid without something specified after the `/` character, so one or more characters would be required.

@JimBugwadia don't know if you saw this second question ☝️

chipzoller (approved):
Looks good. Had one question about wildcards (support for `?*` combined vis-à-vis AppArmor rule).

JimBugwadia:
@chipzoller - regarding the annotation question. Either "?*" or "*" would work for the purposes of this check and "*" seemed simpler and matches what is documented in the [pod security standards](https://kubernetes.io/docs/concepts/security/pod-security-standards/).

PR #7, opened 2020-12-30
update sample policies

JimBugwadia:
- Added titles to policies that require them 
- Updated names and descriptions
- Removed duplicates, etc.


JimBugwadia:
@chipzoller - instructions on rendering policy markdown are available at:

https://github.com/kyverno/website/blob/feature/update_sample_policies/render/README.md

If you have the time, can you please try out and let me know your feedback.

We can potentially automate using a GitHub action or as part of the Netlify site build (not sure if this is necessary at the moment...)

chipzoller:
> @chipzoller - instructions on rendering policy markdown are available at:
> 
> https://github.com/kyverno/website/blob/feature/update_sample_policies/render/README.md
> 
> If you have the time, can you please try out and let me know your feedback.
> 
> We can potentially automate using a GitHub action or as part of the Netlify site build (not sure if this is necessary at the moment...)

@JimBugwadia some feedback:

- Might be good to make this process more "go newbie" friendly such as:
  - Add step to `chmod +x` the `render` binary after `go build` process.
  - May need to change the output of the build process to be something other than `render` if executed in the root of the website dir since `render` already exists as a directory.
  - Link to instructions on how to install go(?)
- When I run render on the current state of the policies repo, it's generating `title` and `linkTitle` with dashes and not with spaces (as the MD files currently have). `weight` is also different. Haven't checked the code.

other/imagepullpolicy-always.yaml
```diff
@@ -12,8 +17,10 @@ spec:
...
     validate:
-      message: "The imagePullPolicy must be set to `Always` for all containers when a tag other than `latest` is used."  
+      message: >-
+        The imagePullPolicy must be set to `Always` when the tag `latest` is used.
       pattern:
         spec:
           containers:
-          - imagePullPolicy: Always
\ No newline at end of file
+          - (image): "*:latest | !*:*"
```
chipzoller, line 25: This expression will only work in Kyverno 1.3.0. For reasons like this, I'm proposing we implement a new annotation (see wiki) which notes the minimum version of Kyverno needed to correctly implement a policy.

pod-security/default/restrict-apparmor-profiles.yaml
```diff
@@ -3,6 +3,7 @@ kind: ClusterPolicy
 metadata:
   name: restrict-apparmor-profiles
   annotations:
+    policies.kyverno.io/title: Restrict AppArmor
```
chipzoller, line 6: This policy is another example which requires Kyverno 1.3.0+.

chipzoller (approved):
Everything looks good. Left a couple of comments regarding suggested new annotation for describing the minimum version of Kyverno needed to implement.

JimBugwadia:
Submitted: https://github.com/kyverno/website/issues/59

I like the idea or adding a minimum version for a policy. We can discuss the expected behaviors, if we want to check / enforce at runtime.

JimBugwadia:
Created: https://github.com/kyverno/website/issues/59

PR #11, opened 2021-01-07
update tests for pod security policies

JimBugwadia:


pod-security/README.md
```diff
@@ -12,4 +12,28 @@ kustomize build https://github.com/kyverno/policies/pod-security | kubectl apply
...
+
+To apply the tests:
+
+1. Install Kyverno (if needed):
+
+```shell
+kubectl create -f https://raw.githubusercontent.com/kyverno/kyverno/main/definitions/release/install.yaml
+```
+
+2. Install pod security policies:
```
chipzoller, line 25: Probably best to not call these "pod security policies" lest we confuse users.

chipzoller (commented):
@JimBugwadia I see that you have converted these test cases over from a Deployment to a bare Pod. The reason I didn't do this is because, while it may produce the same ultimate effect, no one creates standalone Pods; it's always through higher-level abstractions/controllers like Deployments. As a result, this creates two potential problems:

1. it doesn't represent real-world behavior and tests should do that
1. it potentially allows for bugs to slip through Kyverno as testing would be different from actual expected implementations

JimBugwadia:
@chipzoller agree that bare pods should not be used by end users. However, I wanted to make sure the tests cover all controllers and since pods controllers do internally  create pods, that seemed like the best way to write a test. Also, it makes the YAML concise and easier to manage. 

Kyverno does automatically generate rules for the Kubernetes pod controllers so should be possible to add additional tests if needed.

chipzoller:
I'm not sure I fully concur with your reasoning here. When writing a test for a piece of functionality, I don't believe the goal should be to make that test as simple as possible but rather make the test replicate, as accurately as possible, the circumstances the code can expect to see in a "real" situation. Testing against Pods then forces you to guess and assume about a piece of code's ability to operate against different constructs (i.e., "this test worked on a Pod so it **should** work on anything else that creates Pods, right?"). And, although I cannot seem to find it, I'm sure I ran into at least one, maybe two, issues in months past because of this exact type of assumption.

Also, in a way you've actually made the tests more complex as you've taken a Deployment with multiple components and broken those out into separate YAML documents (ex., a single unit test with separate documents, one for initContainer and another for container).

You're welcome to overrule me here and maybe we figure out something else down the line, but I do think it's important that, if we're using tests to reason about how Kyverno works, whether those be "field" tests or unit tests, those tests be just like what users are doing.

JimBugwadia:
@chipzoller I generally agree with what you are saying about testing. 

The goal of these YAMLs is to test allowed configurations on `pod.securityContext` and `pod.containers.securityContext`. Hence, I feel we need to provide inputs at a pod level, and ensure that all controllers are covered.

chipzoller (approved):

PR #29, opened 2021-03-20
replaces description for the require-labels policy

hubwoop:
Hi, the description on the "require-labels" policy was the one for image tags. With this PR an alternative description is suggested. 

The sentences used are copied from https://kubernetes.io/docs/concepts/configuration/overview/#using-labels and https://kubernetes.io/docs/concepts/overview/working-with-objects/common-labels/

Would this be an appropriate description for this policy?


realshuting (approved):
@hubwoop - this looks good to me, thanks for clarifying it!

PR #57, opened 2021-05-10
allow only supplementalGroups greater 0

smnmtzgr:


realshuting:
@smnmtzgr - looks good!

Logged this issue to update tests https://github.com/kyverno/policies/issues/58.

chipzoller:
If the minimum version of Kyverno needed to support this is 1.3.6, we need to add the relevant annotation as seen [here](https://github.com/kyverno/policies/blob/b6aa5dbdca71cf7ac4e0cd09d97b9fc72e067212/pod-security/baseline/restrict-apparmor-profiles/restrict-apparmor-profiles.yaml#L10). This also needs to be held until 1.3.6 releases.

smnmtzgr:
@chipzoller should I add this annotation?

chipzoller:
Yes, let's go ahead and add this annotation now that 1.3.6 is released.

smnmtzgr:
@chipzoller / @realshuting I have added the annotation. Can you check it again and if okay please merge.

pod-security/restricted/require-non-root-groups/require-non-root-groups.yaml
```diff
@@ -5,6 +5,7 @@ metadata:
   annotations:
     policies.kyverno.io/category: Pod Security Standards (Restricted)
     policies.kyverno.io/severity: medium
+    policies.kyverno.io/minversion: 1.6.0
```
realshuting, line 8: Hi @smnmtzgr - the minimum version should be 1.3.6.

realshuting (changes requested):

pod-security/restricted/require-non-root-groups/require-non-root-groups.yaml
smnmtzgr: sorry fixed...

realshuting (approved):

PR #66, opened 2021-06-07
Add 3 new Pod policies

chipzoller:
This PR adds three new sample policies focused on Pods:

* Allowed Pod Priorities
* Require imagePullSecrets
* Limit Containers per Pod

Closes https://github.com/kyverno/kyverno/issues/1968
Closes https://github.com/kyverno/kyverno/issues/1971
Closes https://github.com/kyverno/policies/issues/65
Closes https://github.com/kyverno/kyverno/issues/1929


other/limit_containers_per_pod.yaml
```diff
@@ -0,0 +1,76 @@
+apiVersion: kyverno.io/v1
+kind: ClusterPolicy
+metadata:
+  name: limit-containers-per-pod
+  annotations:
```
realshuting, line 5: Same here.

other/allowed_pod_priorities.yaml
```diff
@@ -0,0 +1,80 @@
+apiVersion: kyverno.io/v1
+kind: ClusterPolicy
+metadata:
+  name: allowed-podpriorities
+  annotations:
```
realshuting, line 5: Since we have all three rules defined, should we disable auto-gen by the annotation? Once https://github.com/kyverno/kyverno/pull/1914 is merged, we can remove the other two rules for pod controllers.

chipzoller: But I thought the auto-gen controller wouldn't apply in those situations because both policies involve `validate.deny` conditions.

chipzoller: When I say "wouldn't apply" I mean even after that PR gets merged. The PR doesn't provide a provision for `validate.deny` conditions, or am I wrong?

realshuting: You are right! I was focus on the path conversion in auto-gen. It looks good!

realshuting (approved):

other/allowed_pod_priorities.yaml
chipzoller: This is why we really need [this functionality](https://github.com/kyverno/kyverno/issues/1933). I think that's the last remaining gap for auto-gen.

PR #70, opened 2021-06-25
Fix capabilities.add of null => "null"

### LINKED ISSUE #69: [BUG] The disallow-adding-capabilities incorrectly specifies null instead of "null"
The `disallow-adding-capabilities` policy uses `null` instead of `"null"`. Because of this, it doesn't actually enforce disallowing adding capabilities and instead I was able to add `NET_ADMIN` like so:

```yaml
apiVersion: apps/v1
kind: Deployment
metadata:
  name: add-caps
spec:
  selector:
    matchLabels:
      app: add-caps
  template:
    metadata:
      labels:
        app: add-caps
    spec:
      containers:
      - name: add-caps
        image: redis
        securityContext:
          capabilities:
            add:
              - "NET_ADMIN"
```

Once I fixed the policy to `"null"` the property was correctly enforced.

When set to `null` and applied, you can note that the rendered pattern spec is:

```yaml
spec:
  rules:
  - name: capabilities
    validate:
      ...
      pattern:
        spec:
          =(initContainers):
          - =(securityContext):
              =(capabilities): {}
          containers:
          - =(securityContext):
              =(capabilities): {}
```

FernFerret:
When `add` is specified as `null` instead of `"null"` the policy is not
actually enforced.

Fixes #69

### DIFF
--- a/pod-security/baseline/disallow-adding-capabilities/disallow-adding-capabilities.yaml
+++ b/pod-security/baseline/disallow-adding-capabilities/disallow-adding-capabilities.yaml
@@ -28,8 +28,8 @@ spec:
             containers:
               - =(securityContext):
                   =(capabilities):
-                    X(add): null
+                    X(add): "null"
             =(initContainers):
               - =(securityContext):
                   =(capabilities):
-                    X(add): null
+                    X(add): "null"


realshuting (approved):
Thanks @FernFerret!

PR #74, opened 2021-07-08
Policy updates

### LINKED ISSUE #68: In policy, kind attribute is missing in resource element
[This](https://github.com/kyverno/policies/blob/main/other/block_updates_deletes.yaml) policy and one of the [examples](https://kyverno.io/docs/writing-policies/match-exclude/#match-a-label-and-exclude-users-and-roles), contains a resource element but not a kind attribute.  Whereas it is mention in the [documentatio](https://kyverno.io/docs/writing-policies/match-exclude/)n that while working with resource elements, the kind attribute is mandatory. So, I think policy needs to be updated with the kind attribute. 
For confirmation, I applied this policy but it returns an error.

`$ kubectl apply -f block-updates-delete.yaml`
`Error from server: error when creating "block-updates-delete.yaml": admission webhook "validate-policy.kyverno.svc" denied the request: At least one element must be specified in a kind block. The kind attribute is mandatory when working with the resources element`

### LINKED ISSUE #71: Require Limits and Requests
I'd like to have a policy that checks for a match between resource requests and limits. Since there doesn't appear to be any kind of == operator, I tried making use of [variables](https://kyverno.io/docs/writing-policies/variables/) like so:

```
            containers:
              - resources:
                  requests:
                    memory: "?*"
                  limits:
                    memory: "$(./../../requests/memory)"
```

To my surprise though, my newly created pod (with non-matching memory requests/limits) passed validation. I suspect the variable is not interpolating what was set for requests.memory and is just using "?*". Am I thinking about this correctly? Is there another method I can use to evaluate?

### LINKED ISSUE #73: re-work ensure_probes_different.yaml
Re-work to see if it can be simplified similar to the lookup here: https://github.com/kyverno/kyverno/issues/2086

chipzoller:
Closes #73
Closes #71 
Closes #68 
Closes https://github.com/kyverno/kyverno/issues/2086
Closes https://github.com/kyverno/kyverno/issues/2067
Closes https://github.com/kyverno/kyverno/issues/1685

### DIFF
--- a/other/ensure_probes_different.yaml
+++ b/other/ensure_probes_different.yaml
@@ -3,70 +3,29 @@ kind: ClusterPolicy
 metadata:
   name: validate-probes  
   annotations:
-    # Only applies to pods originating from DaemonSet, Deployment, or StatefulSet.
-    pod-policies.kyverno.io/autogen-controllers: DaemonSet,Deployment,StatefulSet  
+    pod-policies.kyverno.io/autogen-controllers: none  
     policies.kyverno.io/title: Validate Probes
     policies.kyverno.io/category: Sample
     policies.kyverno.io/severity: medium
+    policies.kyverno.io/minversion: 1.3.6
     policies.kyverno.io/subject: Pod
     policies.kyverno.io/description: >-
-      Sample policy to check that liveness and readiness probes are not set to the same values.
+      Sample policy to check that liveness and readiness probes are not equal.
 spec:
-  validationFailureAction: audit
+  validationFailureAction: enforce
   background: false
   rules:
-    # Checks the first container in a Pod.
-    - name: validate-probes-c0
+    - name: validate-probes
       match:
         resources:
           kinds:
-          - Pod
-      validate:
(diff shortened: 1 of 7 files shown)


other/ensure_probes_different.yaml
```diff
@@ -3,70 +3,29 @@ kind: ClusterPolicy
...
-          - Pod
-      validate:
-        message: "Liveness and readiness probes cannot be the same."
-        deny:
-          conditions:
-          - key: "{{ request.object.spec.containers[3].readinessProbe }}"
-            operator: Equals
-            value: "{{ request.object.spec.containers[3].livenessProbe }}"
\ No newline at end of file
+          - key: "{{ request.object.spec.template.spec.containers[?readinessProbe!=livenessProbe] | length(@) }}"
```
realshuting, line 29: 👍

realshuting (approved):

chipzoller:
@JimBugwadia do you have a moment to provide review for this PR which incorporates several new policies?

JimBugwadia (approved):

Issue #80, opened 2021-07-19
Sample policy to validate for v1.22 deprecated APIs


JimBugwadia:
See: https://twitter.com/Marcus_Noble_/status/1417007780888825856

Kubernetes v1.22 removes a number of deprecated APIs. The policy would check for those apiVersions and kinds and block / report their usage.

JimBugwadia:
Here is a sample:

```yaml
apiVersion: kyverno.io/v1
kind: ClusterPolicy
metadata:
  name: check-deprecated-apis
  annotations:
    policies.kyverno.io/category: Best Practices
    policies.kyverno.io/description: >-
      As the Kubernetes API evolves, APIs are periodically reorganized or upgraded. 
      When APIs evolve, the old APIs they replace are deprecated, and eventually removed. 
      See Kubernetes API removals to read more about Kubernetes' policy on removing APIs.
      See: https://kubernetes.io/blog/2021/07/14/upcoming-changes-in-kubernetes-1-22/ 
      for details.
spec:
  validationFailureAction: enforce
  rules:
  - name: validate-v1-22-removals
    match:
      resources:
        kinds:
        - admissionregistration.k8s.io/v1beta1/ValidatingWebhookConfiguration
        - admissionregistration.k8s.io/v1beta1/MutatingWebhookConfiguration
        - apiextensions.k8s.io/v1beta1/CustomResourceDefinition
        - apiregistration.k8s.io/v1beta1/APIService
        - authentication.k8s.io/v1beta1/TokenReview
        - authorization.k8s.io/v1beta1/SubjectAccessReview
        - authorization.k8s.io/v1beta1/LocalSubjectAccessReview
        - authorization.k8s.io/v1beta1/SelfSubjectAccessReview 
        - certificates.k8s.io/v1beta1/CertificateSigningRequest
        - coordination.k8s.io/v1beta1/Lease
        - extensions/v1beta1/Ingress
        - networking.k8s.io/v1beta1/Ingress
    validate:
      message: "{{ request.object.apiVersion }}/{{ request.object.kind }} is deprecated and will be removed in v1.22. See: https://kubernetes.io/blog/2021/07/14/upcoming-changes-in-kubernetes-1-22/"
      deny: {}
```

When used with this YAML:

```yaml
apiVersion: networking.k8s.io/v1beta1
kind: Ingress
metadata:
  name: ingress-fanout
  annotations:
    nginx.ingress.kubernetes.io/rewrite-target: /
spec:
  rules:
[...]

PR #81, opened 2021-07-19
check_deprecated_apis.yaml

JimBugwadia:
Signed-off-by: Jim Bugwadia <jim@nirmata.com>

Fixes https://github.com/kyverno/policies/issues/80

### DIFF
--- a/best-practices/check_deprecated_apis.yaml
+++ b/best-practices/check_deprecated_apis.yaml
@@ -1,0 +1,33 @@
+apiVersion: kyverno.io/v1
+kind: ClusterPolicy
+metadata:
+  name: check-deprecated-apis
+  annotations:
+    policies.kyverno.io/category: Best Practices
+    policies.kyverno.io/description: >-
+      Kubernetes APIs are sometimes deprecated and removed after a few releases. 
+      As a best practice older API versions should be replaced with newer versions.
+spec:
+  validationFailureAction: enforce
+  rules:
+  - name: validate-v1-22-removals
+    match:
+      resources:
+        kinds:
+        - admissionregistration.k8s.io/v1beta1/ValidatingWebhookConfiguration
+        - admissionregistration.k8s.io/v1beta1/MutatingWebhookConfiguration
+        - apiextensions.k8s.io/v1beta1/CustomResourceDefinition
+        - apiregistration.k8s.io/v1beta1/APIService
+        - authentication.k8s.io/v1beta1/TokenReview
+        - authorization.k8s.io/v1beta1/SubjectAccessReview
+        - authorization.k8s.io/v1beta1/LocalSubjectAccessReview
+        - authorization.k8s.io/v1beta1/SelfSubjectAccessReview 
+        - certificates.k8s.io/v1beta1/CertificateSigningRequest
+        - coordination.k8s.io/v1beta1/Lease
+        - extensions/v1beta1/Ingress
+        - networking.k8s.io/v1beta1/Ingress
+    validate:
+      message: >-
+        {{ request.object.apiVersion }}/{{ request.object.kind }} is deprecated and will be removed in v1.22. 
+        See: https://kubernetes.io/blog/2021/07/14/upcoming-changes-in-kubernetes-1-22/
+      deny: {}
--- a/best-practices/check_deprecated_apis.yaml
+++ b/best-practices/check_deprecated_apis.yaml
@@ -6,9 +6,10 @@ metadata:
     policies.kyverno.io/category: Best Practices
     policies.kyverno.io/description: >-
       Kubernetes APIs are sometimes deprecated and removed after a few releases. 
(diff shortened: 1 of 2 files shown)


best-practices/check_deprecated_apis.yaml
```diff
@@ -0,0 +1,33 @@
...
+kind: ClusterPolicy
+metadata:
+  name: check-deprecated-apis
+  annotations:
+    policies.kyverno.io/category: Best Practices
+    policies.kyverno.io/description: >-
+      Kubernetes APIs are sometimes deprecated and removed after a few releases. 
+      As a best practice older API versions should be replaced with newer versions.
+spec:
+  validationFailureAction: enforce
```
chipzoller, line 11: I'm trying to standardize that all validate policies are set to `audit`. This is to ensure no unintended blocking happens if users want to take a policy for a "spin" and don't quite realize what may happen.

best-practices/check_deprecated_apis.yaml
```diff
@@ -0,0 +1,33 @@
...
+  rules:
+  - name: validate-v1-22-removals
+    match:
+      resources:
+        kinds:
+        - admissionregistration.k8s.io/v1beta1/ValidatingWebhookConfiguration
```
chipzoller, line 17: Did you see [this issue](https://github.com/kyverno/kyverno/issues/2158#issuecomment-882023960)? Not sure this will do anything if accurate.

best-practices/check_deprecated_apis.yaml
```diff
@@ -0,0 +1,33 @@
+apiVersion: kyverno.io/v1
+kind: ClusterPolicy
+metadata:
+  name: check-deprecated-apis
+  annotations:
+    policies.kyverno.io/category: Best Practices
+    policies.kyverno.io/description: >-
+      Kubernetes APIs are sometimes deprecated and removed after a few releases. 
+      As a best practice older API versions should be replaced with newer versions.
```
chipzoller, line 9: Would suggest adding to the description annotation something to the effect of: "This policy validates that no resources using any of these deprecated APIs are created."

best-practices/check_deprecated_apis.yaml
```diff
@@ -0,0 +1,33 @@
...
+      resources:
+        kinds:
+        - admissionregistration.k8s.io/v1beta1/ValidatingWebhookConfiguration
+        - admissionregistration.k8s.io/v1beta1/MutatingWebhookConfiguration
+        - apiextensions.k8s.io/v1beta1/CustomResourceDefinition
+        - apiregistration.k8s.io/v1beta1/APIService
+        - authentication.k8s.io/v1beta1/TokenReview
+        - authorization.k8s.io/v1beta1/SubjectAccessReview
+        - authorization.k8s.io/v1beta1/LocalSubjectAccessReview
+        - authorization.k8s.io/v1beta1/SelfSubjectAccessReview
```
serialoverflow, line 24: These are blocked in the default resource filters: `'[Event,*,*][*,kube-system,*][*,kube-public,*][*,kube-node-lease,*][Node,*,*][APIService,*,*][TokenReview,*,*][SubjectAccessReview,*,*][SelfSubjectAccessReview,*,*][*,kyverno,*] [Binding,*,*][Repl      icaSet,*,*][ReportChangeRequest,*,*][ClusterReportChangeRequest,*,*]'`

chipzoller: You are, of course, correct, but in this case a policy which specifies these may still be useful in cases where users have gone and modified the ConfigMap after the fact (and possibly can't envision the future impact). If they remain filtered out in the ConfigMap, then this policy will obviously not be able to block them.

best-practices/check_deprecated_apis.yaml
JimBugwadia: A violation will still be shown in the policy reports, right?

best-practices/check_deprecated_apis.yaml
JimBugwadia: Sounds good. Updated.

best-practices/check_deprecated_apis.yaml
JimBugwadia: Good point! Wondering if we need to add a warning about policies that use kinds that are filtered, as a Kyverno enhancement.

best-practices/check_deprecated_apis.yaml
chipzoller: Good point, not sure.

best-practices/check_deprecated_apis.yaml
chipzoller: That's a good idea!

JimBugwadia: Added: https://github.com/kyverno/kyverno/issues/2163.

chipzoller (approved):

PR #84, opened 2021-07-22
Fix externalIPs of nil => "null"

vyankyGH:


chipzoller:
Why'd you close this?

PR #86, opened 2021-07-22
fix negation anchor hostPort nil => null

vyankyGH:
Please See: https://github.com/kyverno/kyverno/issues/1820

### DIFF
--- a/pod-security/baseline/disallow-host-ports/disallow-host-ports.yaml
+++ b/pod-security/baseline/disallow-host-ports/disallow-host-ports.yaml
@@ -26,7 +26,8 @@ spec:
           spec:
             =(initContainers):
               - =(ports):
-                  - X(hostPort): "null"
+              # the value here is actually not evaluated, the key is the one do not allow.
+                  - X(hostPort): 0
             containers:
               - =(ports):
-                  - X(hostPort): "null"
+                  - X(hostPort): 0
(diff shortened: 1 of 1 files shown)


pod-security/baseline/disallow-host-ports/disallow-host-ports.yaml
```diff
@@ -26,7 +26,7 @@ spec:
           spec:
             =(initContainers):
               - =(ports):
-                  - X(hostPort): 0
+                  - X(hostPort): "null"
```
JimBugwadia, line 29: Why does this value need to be "null"?

vyankyGH: Added Validation for negation anchor value must be null

chipzoller: I think you might have been confused by that issue. The value is irrelevant in Kyverno's eyes because it only looks at the key. The value of a field using the negation anchor on the key could just as well be undefined. The relevant issue is only for validating the policy itself.

pod-security/baseline/disallow-host-ports/disallow-host-ports.yaml
```diff
@@ -26,7 +26,8 @@ spec:
           spec:
             =(initContainers):
               - =(ports):
-                  - X(hostPort): "null"
+              # the value here is actually not evaluated, the key is the one do not allow.
```
chipzoller, line 29: Comment here really isn't needed since this has been updated in the documentation.

PR #101, opened 2021-08-17
change background to true

### LINKED ISSUE #97: `background: false` is specified even though it is not required.
In some sample policies, `background: false` is specified even though it is not required.  The following are some examples.

- https://github.com/kyverno/policies/blob/main/best-practices/restrict_image_registries/restrict_image_registries.yaml#L18
- https://github.com/kyverno/policies/blob/main/other/imagepullpolicy-always.yaml#L17
- https://github.com/kyverno/policies/blob/main/other/restrict_annotations.yaml#L18
- https://github.com/kyverno/policies/blob/main/other/restrict_controlplane_scheduling.yaml#L17
- https://github.com/kyverno/policies/blob/main/other/restrict_node_selection.yaml#L17

This is confusing and I think we should specify `background: false` only where necessary. Alternatively, we should explicitly specify whether the background attribute is true or false for all policies.

sotoiwa:
Since this policy allows for background scanning, I changed the background attribute true.

Closes #97

### DIFF
--- a/other/block_updates_deletes.yaml
+++ b/other/block_updates_deletes.yaml
@@ -13,7 +13,7 @@ metadata:
       Service resource that contains the label `protected=true` unless by
       a cluster-admin.
 spec:
-  validationFailureAction: audit
+  validationFailureAction: enforce
   background: false
   rules:
   - name: block-updates-deletes
--- a/other/block_updates_deletes.yaml
+++ b/other/block_updates_deletes.yaml
@@ -13,7 +13,7 @@ metadata:
       Service resource that contains the label `protected=true` unless by
       a cluster-admin.
 spec:
-  validationFailureAction: enforce
+  validationFailureAction: audit
   background: false
   rules:
   - name: block-updates-deletes
@@ -23,7 +23,7 @@ spec:
         - Service
         selector:
           matchLabels:
-            protected: "true"   
+            protected: "true"
     exclude:
       clusterRoles:
       - cluster-admin
@@ -33,6 +33,6 @@ spec:
         conditions:
           - key: "{{request.operation}}"
             operator: In
-            value: 
+            value:
             - DELETE
             - UPDATE
--- a/other/verify_image.yaml
+++ b/other/verify_image.yaml
@@ -16,7 +16,7 @@ metadata:
       its signature against the provided public key. This policy serves as an illustration for
       how to configure a similar rule and will require replacing with your image(s) and keys.
 spec:
-  validationFailureAction: audit
+  validationFailureAction: enforce
   background: false
   rules:
     - name: verify-image
--- a/other/verify_image.yaml
+++ b/other/verify_image.yaml
@@ -16,7 +16,7 @@ metadata:
       its signature against the provided public key. This policy serves as an illustration for
       how to configure a similar rule and will require replacing with your image(s) and keys.
 spec:
-  validationFailureAction: enforce
+  validationFailureAction: audit
   background: false
   rules:
     - name: verify-image
--- a/best-practices/add_network_policy.yaml
+++ b/best-practices/add_network_policy.yaml
@@ -7,19 +7,20 @@ metadata:
     policies.kyverno.io/category: Multi-Tenancy
     policies.kyverno.io/subject: NetworkPolicy
     policies.kyverno.io/description: >-
-      By default, Kubernetes allows communications across all Pods within a cluster. 
-      The NetworkPolicy resource and a CNI plug-in that supports NetworkPolicy must be used to restrict 
-      communications. A default NetworkPolicy should be configured for each Namespace to 
-      default deny all ingress and egress traffic to the Pods in the Namespace. Application 
-      teams can then configure additional NetworkPolicy resources to allow desired traffic 
+      By default, Kubernetes allows communications across all Pods within a cluster.
+      The NetworkPolicy resource and a CNI plug-in that supports NetworkPolicy must be used to restrict
+      communications. A default NetworkPolicy should be configured for each Namespace to
+      default deny all ingress and egress traffic to the Pods in the Namespace. Application
+      teams can then configure additional NetworkPolicy resources to allow desired traffic
       to application Pods from select sources. This policy will create a new NetworkPolicy resource
       named `default-deny` which will deny all traffic anytime a new Namespace is created.
 spec:
   validationFailureAction: audit
+  background: false
   rules:
   - name: default-deny
     match:
-      resources: 
+      resources:
         kinds:
         - Namespace
     generate:
@@ -32,6 +33,6 @@ spec:
           # select all pods in the namespace
           podSelector: {}
           # deny all traffic
-          policyTypes: 
+          policyTypes:
           - Ingress
           - Egress
--- a/best-practices/add_network_policy.yaml
+++ b/best-practices/add_network_policy.yaml
@@ -15,7 +15,6 @@ metadata:
       to application Pods from select sources. This policy will create a new NetworkPolicy resource
       named `default-deny` which will deny all traffic anytime a new Namespace is created.
 spec:
-  background: false
   rules:
   - name: default-deny
     match:
--- a/best-practices/add_network_policy.yaml
+++ b/best-practices/add_network_policy.yaml
@@ -15,7 +15,6 @@ metadata:
       to application Pods from select sources. This policy will create a new NetworkPolicy resource
       named `default-deny` which will deny all traffic anytime a new Namespace is created.
 spec:
-  validationFailureAction: audit
   background: false
   rules:
   - name: default-deny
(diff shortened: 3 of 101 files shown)


chipzoller:
If you wanted to take this up, it would be great to lump all these changes into a single PR rather than just a single policy update.

sotoiwa:
@chipzoller Would you please review my PR? The changes I have made are as follows:

- Specified validationFailureAction and background attribute in all policies including mutate and generate.
- `validationFailureAction` is now `audit` for all policies  including mutate and generate.
- `background` is now `false` for the mutate and generate policies.
- For the validate policy, `background` is `true` if possible.
  - I used the `kyverno verify` command to check if the `background` can be set to `true`. 
- Wrote `validationFailureAction` and `background` attributes in this order.
- Removed unnecessary trailing spaces and newlines.
- Some of the yaml was indented with a single space, so I fixed it to indent with two spaces.

chipzoller:
Because `validationFailureAction` is only applicable to `validate` rules, I would recommend eliminating that line from `mutate` and `generate` policies because it's irrelevant, ignored by Kyverno, and bloats (if only slightly) sample policies.

other/verify_image.yaml
```diff
@@ -16,7 +16,7 @@ metadata:
       its signature against the provided public key. This policy serves as an illustration for
       how to configure a similar rule and will require replacing with your image(s) and keys.
 spec:
-  validationFailureAction: enforce
+  validationFailureAction: audit
```
chipzoller, line 19: This one should remain in `enforce` mode because it's pointless to have it in `audit` mode.

other/block_updates_deletes.yaml
```diff
@@ -13,7 +13,7 @@ metadata:
       Service resource that contains the label `protected=true` unless by
       a cluster-admin.
 spec:
-  validationFailureAction: enforce
+  validationFailureAction: audit
```
chipzoller, line 16: This one should remain in `enforce` mode because it's pointless to have it in `audit` mode.

sotoiwa: It seemed to me that other `background: false` validate policies (e.g. block-pod-exec-by-pod-label.yaml) are also useless in `audit` mode.  What do you think?

sotoiwa:
> Because validationFailureAction is only applicable to validate rules, I would recommend eliminating that line from mutate and generate policies because it's irrelevant, ignored by Kyverno, and bloats (if only slightly) sample policies.

I agree with you. I will remove the `validationFailureAction` attribute from all mutate and generate policies.

I think the `background` attribute is also only applicable to validate rules and ignored by kyverno in mutate and generate polices, but I'll leave `background: false` listed to remind people that it won't be background scanned.

other/verify_image.yaml
sotoiwa: It seemed to me that other `background: false` validate policies (e.g. block-pod-exec-by-pod-label.yaml) are also useless in `audit` mode. I don't understand why only this `verify_image.yaml` and the next `block_updates_deletes.yaml` should be in `enforce` mode.

chipzoller: The samples that block operations are not really valuable in background mode, so setting those to `false` makes sense. And same thing with these being in `audit` mode. If the whole point of the policy is to block updates and deletes, putting in audit does nothing. That's the exception. With the verify image policy, same sort of deal except that policy is really only an illustration, and no user is actually going to use that policy as-is without modification.

other/block_updates_deletes.yaml
chipzoller: The exec policies should probably be in `enforce` mode, but having those in background mode would help users get a sense if a Pod was exec'd into to assess the value of changing such a policy to `enforce`.

chipzoller:
> 
> 
> > Because validationFailureAction is only applicable to validate rules, I would recommend eliminating that line from mutate and generate policies because it's irrelevant, ignored by Kyverno, and bloats (if only slightly) sample policies.
> 
> I agree with you. I will remove the `validationFailureAction` attribute from all mutate and generate policies.
> 
> I think the `background` attribute is also only applicable to validate rules and ignored by kyverno in mutate and generate polices, but I'll leave `background: false` listed to remind people that it won't be background scanned.

I'd suggestion eliminating it entirely. Otherwise, users may be under the false impression that if they change `false` to `true` on such policies that they'll work retroactively, and that's not how Kyverno works (at present, anyway).

other/block_updates_deletes.yaml
sotoiwa: The block-pod-exec-* policy uses `request.operation`, so it can't be `background: true`. Should I still leave it as `audit` in this case?

block_updates_deletes.yaml and block-pod-exec-* are very similar rules, so it seems strange that these rules have different values for `validationFailureAction`.

other/verify_image.yaml
sotoiwa: Does the verifyImages rule allow background scan? I couldn't find any mention of it in the documentation. If background scan is possible, I thought it would be better to specify `background: true`.

sotoiwa:
> I'd suggestion eliminating it entirely. Otherwise, users may be under the false impression that if they change false to true on such policies that they'll work retroactively, and that's not how Kyverno works (at present, anyway).

Thank you for the suggestion. I eliminated both `validationFailureAction` and `background` attribute from all mutate and generate policies.

other/block_updates_deletes.yaml
chipzoller: Any of the five block policies which use the CONNECT feature probably need to have `background: false`.

other/verify_image.yaml
chipzoller: Kyverno v1.4.2 allows one to create this policy with `background` in either mode. @JimBugwadia what is supported here officially? I'm guessing `background: true` doesn't actually do anything. Can you confirm? If that ends up being the case, shouldn't we prohibit such policies from being created/updated?

chipzoller:
Once all questions/conversations are resolved, indicate when this PR is ready for final review.

other/block_updates_deletes.yaml
sotoiwa: Sorry, my point was not clear.

Those pod-exec-* policies are already `background: false` and I have not changed `background`.

Those pod-exec-* policies have `validationFailureAction: audit` which I have not changed also. On the other hand, block_updates_deletes.yaml has `validationFailureAction: enforce`, and I changed it to `validationFailureAction: audit` because I think pod-exec-* and block_updates are same kind of rule.

If we leave `validationFailureAction: enforce` of block_updates as it is, then I think other policies that require `background: false` because of `requests.operation`, such as block-pod-exec-*, should also have `validationFailureAction: enforce`.

chipzoller: Ok, let's standardize on policies that validate for non-CREATE operations (UPDATE, CONNECT, DELETE) should be set `validationFailureAction: enforce`. Background mode should be set accordingly to `false`.

sotoiwa: Thank you for the comment. I changed policies that validate for non-CREATE operations (UPDATE, CONNECT, DELETE) to have validationFailureAction: enforce.

other/verify_image.yaml
sotoiwa: I have changed the `validationFailureAction` back to `enforce`, and the `background` is `false` as in the original.

chipzoller:
Is this ready for final review, @sotoiwa ?

sotoiwa:
@chipzoller Yes, it's ready for review. Thanks.

sotoiwa:
@chipzoller Would you please review the PR?

chipzoller:
Apologies. Looks like we may need to hold this because the test file result will be changing with a commit that was made recently. Rather than merge this PR and go back to update all of them again, we might just wait until 1.4.3 is cut and update the files.

chipzoller:
@vyankyGH can we accept this in 1.4.3 now?

chipzoller:
@viveksahu26 see above

vyankyGH:
> @vyankyGH can we accept this in 1.4.3 now?

Let me go through all the changes. They are too many files.

> @vyankyGH can we accept this in 1.4.3 now?

Yes I have tested with 1.4.3. Work's fine.

vyankyGH:
> @viveksahu26 see above

@viveksahu26 can you please test with current test command mutate changes.
**kyverno test https://github.com/sotoiwa/policies/patch-1**

viveksahu26:
> 
> 
> > @viveksahu26 see above
> 
> @viveksahu26 can you please test with current test command mutate changes.
> **kyverno test https://github.com/sotoiwa/policies/patch-1**

All of them are fails, as it doesn't has kind field.
image


image

chipzoller:
@sotoiwa would you be able to add the `kind` field to each of the test cases, please? This is a new requirement as of 1.4.3.

sotoiwa:
@chipzoller I could rebased the branch with no conflicts. So all the tests should now have `kind` field.

chipzoller (approved):

viveksahu26:
Hey @chipzoller , one question. Is `validationActionFailue: true`  for all validate policy.

chipzoller:
Most validate policies should be set to `audit` except for the ones that make no sense to be like Pod exec policies.

PR #106, opened 2021-08-20
Fix the result for restrict-image-registries

realshuting:
According to this policy https://github.com/kyverno/policies/blob/ef49652730b807f70d483dc74ef52fcff8a41194/best-practices/restrict_image_registries/restrict_image_registries.yaml#L31

The result should be `fail` as the image is from `ghcr.io`.


chipzoller:
Thanks for catching this.

PR #108, opened 2021-08-21
Policy updates and fixes

chipzoller:
Closes #102
Closes #109 
Closes #110 
Closes #96 
Closes #124 

This PR updates the PSS restricted policy "restrict-seccomp" to comply with the latest published standard and adds much more comprehensive tests.

It also updates the message Kyverno prints in the "require drop all" policy to indicate that `ALL` must be specified in all caps.

PR #118, opened 2021-09-02
Removed obsolete condition which caused a deny in any case

### LINKED ISSUE #116: Unique Ingress Host sample policy will always validate/deny
The precondition requires any of two opposite checks which will always be true:

```
any:
  - key: "{{ request.object.spec.rules[].host }}"
    operator: In
    value: "{{ hosts }}"
  - key: "{{ request.object.spec.rules[].host }}"
    operator: NotIn
    value: "{{ hosts }}"
```

As we have a deny rule here, the second condition is the one to be removed.

thorstenhuhn:
Closes #116

### DIFF
--- a/other/restrict_ingress_host.yaml
+++ b/other/restrict_ingress_host.yaml
@@ -35,9 +35,6 @@ spec:
         - key: "{{ request.object.spec.rules[].host }}"
           operator: In
           value: "{{ hosts }}"
-        - key: "{{ request.object.spec.rules[].host }}"
-          operator: NotIn
-          value: "{{ hosts }}"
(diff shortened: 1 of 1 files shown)


chipzoller:
I think if we can get confirmation [here ](https://github.com/kyverno/policies/issues/116#issuecomment-922117935) then this PR can safely be accepted.

chipzoller:
Incorporating this fix into a modified rule per the last conversation [here](https://github.com/kyverno/policies/pull/108/commits/4cb606b7955ac47723e256df71f5f23cb5fa1bda).

PR #127, opened 2021-09-16
policy helm chart initial seed

krisek:
This is the initial revision of a collection of Kyverno policies that can be installed with Helm.


krisek:
Apologies for the misunderstanding, it contained a file it shouldn't have.

PR #141, opened 2021-10-04
Add mutation policy for terminationGracePeriodSeconds

leo8a:
Pods with large `terminationGracePeriodSeconds` (tGPS) might prevent cluster nodes from getting drained, ultimately making the system unstable. This policy mutates all incoming Pods to set their tGPS under 50s. If the user creates a pod without specifying tGPS, then the Kubernetes default of 30s is maintained.


realshuting (commented):
Thanks @leo8a for the contribution, it looks good! 

@chipzoller - do we need to add version annotation for this policy?

PR #181, opened 2021-11-30
add ephemeralContainers to all policies

chrisns:
## Related issue
none
## Milestone of this PR
n/a
## What type of PR is this

/kind feature

## Proposed Changes

Added ephemeralContainers to policies where they aren't already

## Checklist

- [x] I have read the [contributing guidelines](https://github.com/kyverno/kyverno/blob/main/CONTRIBUTING.md).
- [x] I have added tests that prove my fix is effective or that my feature works.
- [ ] My PR contains new or altered behavior to Kyverno and
- [x] I have read the [PR documentation guide](https://github.com/kyverno/kyverno/blob/main/.github/pr_documentation.md) and followed the process including adding proof manifests to this PR.

### DIFF
--- a/best-practices/disallow_latest_tag/disallow_latest_tag.yaml
+++ b/best-practices/disallow_latest_tag/disallow_latest_tag.yaml
@@ -27,6 +27,10 @@ spec:
         spec:
           containers:
           - image: "*:*"
+          =(initContainers):
+          - image: "*:*"
+          =(ephemeralContainers):
+          - image: "*:*"
   - name: validate-image-tag
     match:
       resources:
@@ -37,4 +41,8 @@ spec:
       pattern:
         spec:
           containers:
+          - image: "!*:latest"
+          =(initContainers):
+          - image: "!*:latest"
+          =(ephemeralContainers):
           - image: "!*:latest"
--- a/best-practices/require_ro_rootfs/require_ro_rootfs.yaml
+++ b/best-practices/require_ro_rootfs/require_ro_rootfs.yaml
@@ -29,3 +29,9 @@ spec:
           containers:
           - securityContext:
               readOnlyRootFilesystem: true
+          =(ephemeralContainers):
+          - securityContext:
+              readOnlyRootFilesystem: true
+          =(initContainers):
+          - securityContext:
+              readOnlyRootFilesystem: true
--- a/best-practices/restrict_image_registries/restrict_image_registries.yaml
+++ b/best-practices/restrict_image_registries/restrict_image_registries.yaml
@@ -29,3 +29,7 @@ spec:
         spec:
           containers:
           - image: "eu.foo.io/* | bar.io/*"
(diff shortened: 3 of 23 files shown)


realshuting:
@chipzoller - can you please help review?

chipzoller:
I'm going to have to think on this one. On the one hand, it's good to include these (and I did think of covering them in the past but ephemeral containers is presently still behind a feature gate), but on the other unless each one of these has been tested and confirmed to work (and the test resources have been included to verify) we really try not to change the policies.

chrisns:
@chipzoller can add to more to more test cases
Though if theres a more repeatable way of articulating `containers|=(initcontainers)|=(ephemeralcontainers)` rather than copy pasting maybe this would be less of an issue (though a different PR to this!)

The current status quo is theres a lot of policy that talks about `initContainers` but the tests often only cover `containers`, I'm keen to not make that problem worse!

Looking forward to your thoughts 😄

chipzoller:
I think we need to prove that the additional modifications do in fact work rather than a copy-and-assume approach. Any help you can offer to cover these with test cases (or just show with some Proof Manifests in individual PRs) would be great. Because that's going to be a lot of work, I would suggest closing this PR for now and breaking these out into individual PRs per policy which gives you time to test on your side and write more test cases.

chipzoller (commented):
I think this is too large and hasn't been tested to be generally accepted at this time. The suggestion is to split these out into individual PRs which gives time for testing and further test case development.

raesene:
One thing to note on this is, as of Kubernetes 1.23 ephemeralContainers has progressed to beta and is enabled by default.

JimBugwadia:
@chipzoller - any suggestions on how to decompose and tackle? 

[PSS ](https://kubernetes.io/docs/concepts/security/pod-security-standards/#baseline) also includes `ephemeralContainers` so we will need to update accordingly.

chrisns:
in hindsight my pr is probably a fair bit wrong in various places, not all of the fields exist on ephemeral containers, so need to check the actual pod spec. checking won't do any harm obviously, but there is some redundancy

chipzoller:
I'm aware that ephemeralContainers are now on by default, and we do need to address this in the policies. It's just going to take some time to work through them individually and ensure they work as intended as opposed to guessing about their operation based on what may already exist. I don't see another way here than re-touching each policy and verifying the changes have the expected result.

chipzoller:
Also, none of this really matters until Kyverno actually supports ephemeralContainers: https://github.com/kyverno/kyverno/issues/2821

chipzoller:
I've created a temporary repo [here ](https://github.com/chipzoller/kyverno-pss)which includes the current "refresh" of the upstream Pod Security Standard policies for Kyverno and these also address the ephemeral containers. I'll go ahead and close this out but will refer back to it as we keep moving through other pre-existing policies which need ephemeral containers added.

PR #186, opened 2021-12-03
Revert "temporarily remove failing tests"

chrisns:
Reverts kyverno/policies#185
can be merged when 1.6.0 is available (and these tests past)


chipzoller:
These files already got renamed to the new `kyverno-test.yaml` standard, and since the CI workflow now tests on the CLI from kyverno/kyverno:main, I'm closing this as it looks like we're good.

PR #245, opened 2022-01-31
fix: forget to add match keyword

developer-guy:
feat: add two new verify image policies

## Description

We forgot to add match: keyword.

cc: @chipzoller

## Checklist


chipzoller (approved):
Good catch, thanks

PR #246, opened 2022-01-31
feat: add runAsNonRoot:true to default

developer-guy:

## Description

I noticed that we have a check for runAsNonRoot in validation rules but we don't have any mutation for that.

This PR will fix that.

kindly ping @chipzoller

Does it make sense? 

## Checklist

### DIFF
--- a/other/add-default-securitycontext/add-default-securitycontext.yaml
+++ b/other/add-default-securitycontext/add-default-securitycontext.yaml
@@ -9,7 +9,7 @@ metadata:
     policies.kyverno.io/description: >-
       A Pod securityContext entry defines fields such as the user and group which should be used to run the Pod.
       Sometimes choosing default values for users rather than blocking is a better alternative to not impede
-      such Pod definitions. This policy will mutate a Pod to set `runAsUser`, `runAsGroup`, and `fsGroup` fields
+      such Pod definitions. This policy will mutate a Pod to set `runAsNonRoot`, runAsUser`, `runAsGroup`, and `fsGroup` fields
       within the Pod securityContext if they are not already set.
 spec:
   rules:
@@ -22,6 +22,7 @@ spec:
       patchStrategicMerge:
         spec:
           securityContext:
+            +(runAsNonRoot): true
             +(runAsUser): 1000
             +(runAsGroup): 3000
             +(fsGroup): 2000
--- a/other/add-default-securitycontext/patchedResource.yaml
+++ b/other/add-default-securitycontext/patchedResource.yaml
@@ -9,6 +9,7 @@ spec:
   - name: nginx
     image: nginx:latest
   securityContext:
+    runAsNonRoot: true
     runAsUser: 1000
     runAsGroup: 3000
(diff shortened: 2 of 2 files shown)


other/add-default-securitycontext/add-default-securitycontext.yaml
```diff
@@ -22,6 +22,7 @@ spec:
       patchStrategicMerge:
         spec:
           securityContext:
+            +(runAsNonRoot): true
```
chipzoller, line 25: Have you tested this that it works properly without setting to `"true"`? The spirit of this is good, just need to make sure it works as written (boolean) and doesn't require string.

chipzoller (commented):
Before approving, need to test and validate on at least the current 1.5.x release and preferably also 1.6-dev-latest.

other/add-default-securitycontext/add-default-securitycontext.yaml
developer-guy: I've updated it as "true" 🙋🏻‍♂️

chipzoller: So does it work?

developer-guy:
> Before approving, need to test and validate on at least the current 1.5.x release and preferably also 1.6-dev-latest.

I've tested it on locally and it worked as expected 👇 

It should be a boolean btw, I got an error when I change it as string
> kyverno-568449c9d6-fdptz kyverno E0131 16:19:10.359522       1 mutation.go:86] WebhookServer "msg"="mutate error" "error"="failed to validate resource mutated by policy add-default-securitycontext: ValidationError(io.k8s.api.core.v
1.Pod.spec.securityContext.runAsNonRoot): invalid type for io.k8s.api.core.v1.PodSecurityContext.runAsNonRoot: got \"string\", expected \"boolean\"" "action"="mutate" "gvk"="/v1, Kind=Pod" "operation"="CREATE" "resource"="default/Pod/"


Screen Shot 2022-01-31 at 19 14 44
Screen Shot 2022-01-31 at 19 22 31

other/add-default-securitycontext/add-default-securitycontext.yaml
developer-guy: It should be a boolean btw, I got an error when I change it as string

> kyverno-568449c9d6-fdptz kyverno E0131 16:19:10.359522 1 mutation.go:86] WebhookServer "msg"="mutate error" "error"="failed to validate resource mutated by policy add-default-securitycontext: ValidationError(io.k8s.api.core.v
1.Pod.spec.securityContext.runAsNonRoot): invalid type for io.k8s.api.core.v1.PodSecurityContext.runAsNonRoot: got "string", expected "boolean"" "action"="mutate" "gvk"="/v1, Kind=Pod" "operation"="CREATE" "resource"="default/Pod/"

chipzoller: Ok

chipzoller (approved):
Thanks for the addition here!

PR #267, opened 2022-02-10
Added resource and test for add-safe-to-evict policy

afzal442:
Signed-off-by: afzal442 <afzal442@gmail.com>

## Description
This PR adds resource and kyverno-test for add-safe-to-evict policy

```go
Executing add-safe-to-evict...

skipped mutate policy add-safe-to-evict -> resource default/Pod/myapp-pod
skipped mutate policy add-safe-to-evict -> resource default/Pod/myapp-pod
skipped mutate policy add-safe-to-evict -> resource default/Pod/myapp-pod
skipped mutate policy add-safe-to-evict -> resource default/Pod/myapp-pod
│───│───────────────────│────────────────────│───────────────────────│────────│
│ # │ POLICY            │ RULE               │ RESOURCE              │ RESULT │
│───│───────────────────│────────────────────│───────────────────────│────────│
│ 1 │ add-safe-to-evict │ annotate-empty-dir │ default/Pod/myapp-pod │ Pass   │
│ 2 │ add-safe-to-evict │ annotate-host-path │ default/Pod/myapp-pod │ Pass   │
│───│───────────────────│────────────────────│───────────────────────│────────│
```
Test Summary: 2 tests passed and 0 tests failed
## Checklist


chipzoller:
Thanks for this contribution. Couple requests:

1. We must remove the old copy of this policy if it's going to be moved to a subfolder. Otherwise it will result in duplication.
2. Your tests only cover skip cases, but it would be more ideal to also add pass cases where the mutation is performed and compare to a patched resource.

best-practices/add_safe_to_evict/resource.yaml
```diff
@@ -0,0 +1,36 @@
...
+  volumes:
+    - name: mydir
+      emptyDir: {}
+
+---
+
+apiVersion: v1
+kind: Pod
+metadata:
+  name: myapp-pod
```
chipzoller, line 23: This resource is the same name as the above. They must be named uniquely and the names updated in the test manifest.

chipzoller (changes requested):
Resources must be uniquely named.

best-practices/add_safe_to_evict/resource.yaml
chipzoller: I'll just fix this myself.

chipzoller (approved):
Fixes needed which I'll supply.

PR #274, opened 2022-02-24
add: new policy limit config map for serviceaccount

eminaktas:
Signed-off-by: eminaktas <eminaktas34@gmail.com>
Signed-off-by: Emin Aktas <emin.aktas@trendyol.com>

## Related Issue(s)

None

## Description

Adds a new policy for limiting a ConfigMap.

```bash
~ kubectl kyverno test other/limit-configmap-for-sa             

Executing limit-configmap-for-sa...

│───│────────────────────────│──────────────────────────────────│───────────────────────────────────────────│────────│
│ # │ POLICY                 │ RULE                             │ RESOURCE                                  │ RESULT │
│───│────────────────────────│──────────────────────────────────│───────────────────────────────────────────│────────│
│ 1 │ limit-configmap-for-sa │ limit-configmap-for-sa-developer │ default/ConfigMap/any-configmap-name-good │ Pass   │
│ 2 │ limit-configmap-for-sa │ limit-configmap-for-sa-developer │ default/ConfigMap/any-configmap-name-bad  │ Pass   │
│───│────────────────────────│──────────────────────────────────│───────────────────────────────────────────│────────│
```

## Checklist

### DIFF
--- a/other/limit-configmap-for-sa/limit_configmap_for_sa.yaml
+++ b/other/limit-configmap-for-sa/limit_configmap_for_sa.yaml
@@ -1,0 +1,55 @@
+apiVersion: kyverno.io/v1
+kind: ClusterPolicy
+metadata:
+  name: limit-configmap-for-sa
+  annotations:
+    policies.kyverno.io/title: Limit ConfigMap to ServiceAccounts for a User
+    policies.kyverno.io/category: Other
+    policies.kyverno.io/severity: medium
+    kyverno.io/kyverno-version: 1.6.0
+    kyverno.io/kubernetes-version: "1.20-1.23"
+    policies.kyverno.io/subject: ConfigMap, ServiceAccount
+    policies.kyverno.io/description: This policy shows how to restrict certain operations on specific ConfigMaps by ServiceAccounts.
+spec:
+  background: false
+  validationFailureAction: audit
+  rules:
+  - name: limit-configmap-for-sa-developer
+    match:
+      any:
+      - resources:
+          kinds:
+          - "ConfigMap"
+        subjects:
+        - kind: ServiceAccount
+          name: developer
+          namespace: kube-system
+      - resources:
+          kinds:
+          - "ConfigMap"
+        subjects:
+        - kind: ServiceAccount
+          name: another-developer
+          namespace: another-namespace
+    preconditions:
+      all:
+      - key: "{{request.object.metadata.namespace}}"
+        operator: In
+        value:
+        - "any-namespace"
+        - "another-namespace"
+      - key: "{{request.object.metadata.name}}"
+        operator: In
+        value:
+        - "any-configmap-name-good"
+        - "another-configmap-name"
+    validate:
+      message: "{{request.object.metadata.namespace}}/{{request.object.kind}}/{{request.object.metadata.name}} resource is protected. Admin or allowed users can change the resource"
+      deny:
+        conditions:
+          all:
+          - key: "{{request.operation}}"
+            operator: "In"
+            value:
+            - "UPDATE"
+            - "CREATE"
--- a/other/limit-configmap-for-sa/kyverno-test.yaml
+++ b/other/limit-configmap-for-sa/kyverno-test.yaml
@@ -1,0 +1,17 @@
+name: limit-configmap-for-sa
+policies:
+  - limit_configmap_for_sa.yaml
+resources:
+  - resource.yaml
+variables: variables.yaml
+results:
+  - policy: limit-configmap-for-sa
+    rule: limit-configmap-for-sa-developer
+    resource: any-configmap-name-good
+    kind: ConfigMap
+    result: fail
+  - policy: limit-configmap-for-sa
+    rule: limit-configmap-for-sa-developer
+    resource: any-configmap-name-bad
+    kind: ConfigMap
+    result: skip
--- a/other/limit-configmap-for-sa/resource.yaml
+++ b/other/limit-configmap-for-sa/resource.yaml
@@ -1,0 +1,15 @@
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: any-configmap-name-good
+  namespace: any-namespace
+data:
+  key: value
+---
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: any-configmap-name-bad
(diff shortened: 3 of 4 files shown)


eminaktas:
cc @chipzoller @samj1912 @realshuting

chipzoller:
Thanks for this PR! Would you mind explain what use case this solves for you? It appears that it is fairly closely gated (single ConfigMap by name, single ServiceAccount by name, in a single Namespace by name) and so we're curious how this would be consumed.

eminaktas:
Yes, that is the case. My motivation is to protect a ConfigMap since users with `developer` ServiceAccount are allowed to take any actions on ConfigMaps in our environment.

Also, we can extend it with more CofigMap, ServiceAccounts and etc. Should I include the below example with another rule? 

```yaml
    match:
      any:
      - resources:
          kinds:
          - "ConfigMap"
        subjects:
        - kind: ServiceAccount
          name: developer
          namespace: kube-system
      - resources:
          kinds:
          - "ConfigMap"
        subjects:
        - kind: ServiceAccount
          name: another-service-account
          namespace: another-namespace
    preconditions:
      all:
      - key: "{{request.object.metadata.namespace}}"
        operator: In
        value:
        - "any-namespace"
        - "another-namespace"
      - key: "{{request.object.metadata.name}}"
        operator: In
        value:
        - "any-configmap-name-good"
        - "another-namespace"
```

chipzoller:
The above should work.

eminaktas:
@chipzoller Updated it. Thanks for the review. 🚀

other/limit-configmap-for-sa/limit_configmap_for_sa.yaml
```diff
@@ -0,0 +1,54 @@
...
+      message: "{{request.object.metadata.namespace}}/{{request.object.kind}}/{{request.object.metadata.name}} resource is protected. Admin or allowed users can change the resource"
+      deny:
+        conditions:
+          all:
+          - key: "{{request.operation}}"
+            operator: "In"
+            value:
+            - "UPDATE"
+            - "CREATE"
+            - "CONNECT"
```
chipzoller, line 54: The `CONNECT` operation is not valid for a ConfigMap. This only applies to Pods when exec'ing.

chipzoller (changes requested):
Please also [annotate your sample](https://github.com/kyverno/policies/wiki/Kyverno-annotations) with additional annotations `kyverno-version` and `kubernetes-version`.

other/limit-configmap-for-sa/limit_configmap_for_sa.yaml
```diff
@@ -0,0 +1,54 @@
+apiVersion: kyverno.io/v1
+kind: ClusterPolicy
+metadata:
+  name: limit-configmap-for-sa
+  annotations:
+    policies.kyverno.io/title: Limit ConfigMap for a User
+    policies.kyverno.io/category: Other
+    policies.kyverno.io/severity: medium
+    policies.kyverno.io/subject: ConfigMap
```
chipzoller, line 9: ConfigMap, ServiceAccount

other/limit-configmap-for-sa/limit_configmap_for_sa.yaml
```diff
@@ -0,0 +1,54 @@
+apiVersion: kyverno.io/v1
+kind: ClusterPolicy
+metadata:
+  name: limit-configmap-for-sa
+  annotations:
+    policies.kyverno.io/title: Limit ConfigMap for a User
```
chipzoller, line 6: Limit ConfigMaps to ServiceAccounts

other/limit-configmap-for-sa/limit_configmap_for_sa.yaml
```diff
@@ -0,0 +1,54 @@
...
+    policies.kyverno.io/category: Other
+    policies.kyverno.io/severity: medium
+    policies.kyverno.io/subject: ConfigMap
+    policies.kyverno.io/description: Limits given ConfigMap for a specific users.
```
chipzoller, line 10: This policy shows how to restrict certain operations on specific ConfigMaps by ServiceAccounts.

other/limit-configmap-for-sa/limit_configmap_for_sa.yaml
eminaktas: Without `CONNECT`, the user can edit the ConfigMap with `kubectl edit cm <resource-name>` or `kubectl apply -f configmap.yaml` if resource exists.

I don't it is intentionally done but it is needed to prevent users to edit ConfigMap. `UPDATE` doesn't cover `PATCH` requestes.

chipzoller: > Without CONNECT, the user can edit the ConfigMap with kubectl edit cm <resource-name> or kubectl apply -f configmap.yaml if resource exists.

I don't believe that's correct. I actually just tried this and either of those two actions results in an `UPDATE` operation which would be blocked by the policy.

other/limit-configmap-for-sa/limit_configmap_for_sa.yaml
```diff
@@ -0,0 +1,56 @@
+apiVersion: kyverno.io/v1
+kind: ClusterPolicy
+metadata:
+  name: limit-configmap-for-sa
+  annotations:
+    policies.kyverno.io/title: Limit ConfigMap to ServiceAccounts for a User
+    policies.kyverno.io/category: Other
+    policies.kyverno.io/severity: medium
+    kyverno.io/kyverno-version: 1.6.0
+    kyverno.io/kubernetes-version: "1.20-1.23"
```
chipzoller, line 10: Did you personally test this with these four versions? If so, that's great. If not, this would need to be done for the annotation's value to be written as such.

eminaktas: Yes, I did. :)

other/limit-configmap-for-sa/limit_configmap_for_sa.yaml
eminaktas: I tried it again now it worked. I think I missed something at that time. I removed the `CONNECT` part. 👍

chipzoller (approved):
Thank you!

PR #279, opened 2022-03-02
Add OpenShift policies

patelrit:
## Checklist


JimBugwadia (approved):

chipzoller (changes requested):
As we are now trying to standardize on a few new annotations described [here](https://github.com/kyverno/policies/wiki/Kyverno-annotations), it would be appreciated if these policies can be marked with `kyverno.io/kyverno-version` and `kyverno.io/kubernetes-version`.

The `policies.kyverno.io/subject` is set to, if necessary, a comma-separate string of the resources in the match clause.

`policies.kyverno.io/category` here should always be `OpenShift`.

patelrit:
Sure, will make the changes.

It would be great if these recommendations for annotations are reported by the kyverno test utility.

chipzoller (changes requested):
Thank you, @patelrit , however I see that the last line from the [comment here](https://github.com/kyverno/policies/pull/279#pullrequestreview-898927362) doesn't appear to be fully addressed. Going forward for all new policies, because non-`any`|`all` `match`|`exclude` statements are deprecated, we'd like to ensure these blocks have one or the other even if it's just a single condition.

chipzoller (approved):
Thank you, @patelrit !

PR #288, opened 2022-03-14
Add policy for ensuring production deployments matches staging env

hoerup:
## Description

Having two environments (production,staging) with namespaces for each, 
make sure that
* every deployment in production has a corresponding deployment in staging
* that a production deployment uses same image name as its staging counterpart
* that a production deployment uses an older or equal image version as its staging counterpart

## Checklist

### DIFF
--- a/other/ensure-production-matches-staging.yml
+++ b/other/ensure-production-matches-staging.yml
@@ -57,13 +57,20 @@ spec:
         kinds:
         - Deployment
     preconditions:
-      any:
+      all:
       - key: "{{request.operation}}"
         operator: In
         value:
         - CREATE        
         - UPDATE
+      - key: "{{ deployment_count }}"
+        operator: GreaterThan
+        value: 0
     context:
+    - name: deployment_count
+      apiCall:
+        urlPath: "/apis/apps/v1/namespaces/staging/deployments"
+        jmesPath: "items[? metadata.name == '{{ request.object.metadata.name}}'  ]  | length(@)"
     - name: deployment_image
       apiCall:
         urlPath: "/apis/apps/v1/namespaces/staging/deployments"
@@ -79,7 +86,7 @@ spec:
 
 
 
-  - name: require-same-or-older-imageversiom
+  - name: require-same-or-older-imageversion
     match:
       resources:
         namespaces:
@@ -87,22 +94,30 @@ spec:
         kinds:
         - Deployment
     preconditions:
-      any:
+      all:
       - key: "{{request.operation}}"
         operator: In
         value:
         - CREATE        
         - UPDATE
+      - key: "{{ deployment_count }}"
+        operator: GreaterThan
+        value: 0
     context:
+    - name: deployment_count
+      apiCall:
+        urlPath: "/apis/apps/v1/namespaces/staging/deployments"
+        jmesPath: "items[? metadata.name == '{{ request.object.metadata.name}}'  ]  | length(@)"
     - name: deployment_ver
       apiCall:
         urlPath: "/apis/apps/v1/namespaces/staging/deployments"
         jmesPath: "items[? metadata.name == '{{ request.object.metadata.name}}'  ] | [0].spec.template.spec.containers[0].image || ':' | split(@, ':') | [1]"
     validate:
-      message: "Every Deployment in production is required to use the same image as in staging"
+      message: "Every Deployment in production is required to use an image version less than or equal to the one in staging"
       deny:
         conditions:
           any:
           - key: "{{ request.object.spec.template.spec.containers[0].image  | split(@, ':') | [1] }}"
             operator: GreaterThan
             value: "{{ deployment_ver }}"            
+
--- a/other/ensure-production-matches-staging.yml
+++ b/other/ensure-production-matches-staging.yml
@@ -1,0 +1,108 @@
+apiVersion: kyverno.io/v1
+kind: ClusterPolicy
+metadata:
+  name: ensure-production-matches-staging
+  annotations:
+    policies.kyverno.io/title: Ensure Production matches staging
+    policies.kyverno.io/category: Sample
+    policies.kyverno.io/minversion: 1.6.0
+    policies.kyverno.io/subject: Deployment
+    policies.kyverno.io/description: >-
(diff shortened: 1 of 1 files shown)


chipzoller:
Thank you for this PR, @hoerup . Based on your rules, I believe I can offer some enhancements that would bolster this a bit. Let me work this up and I'll share some suggestions here.

hoerup:
thanks @chipzoller 
it's my first complex policy and there are several things im unsure of so i guess theres plenty of room for improvement :)

btw is it true that there is currently no way to express a dependency between the rules within a policy?
in this particular case there is imho no reason to check rule no 2 and 3 if the first one fails

chipzoller:
You could use a precondition in rules 2 and 3 which essentially contains rule 1.

chipzoller:
In a couple of the rules, you are hard coding the index `0` for performing an image match. You should change those rules types to use `foreach` so that every image in both the `containers[]` and `initContainers[]` arrays are validated. If any one is not, it should deny the resource.

hoerup:
ok i think i understand 

first a new rule  check to verify that the number of containers and initcontainers in prod is equals the staging

then for the image checks
use foreach to loop over container in request.object.spec.templatre.spec.containers
and within the loop use the loop apiCall  to lookup the container matching element.name
?

chipzoller:
Yes, that sounds about right. Also, your third rule assumes that the tags being used for images follow the semver strategy. If any of them don't that check won't evaluate properly.

chipzoller:
Any updates here? Also note that DCO sign-off is required.

chipzoller:
Thank you for this PR. I have extended it myself to do the tasks you intended and will include this in the next round of policy updates. Closing.

```yaml
apiVersion: kyverno.io/v1
kind: ClusterPolicy
metadata:
  name: ensure-production-matches-staging
  annotations:
    policies.kyverno.io/title: Ensure Production Matches staging
    policies.kyverno.io/category: Other
    policies.kyverno.io/minversion: 1.6.0
    kyverno.io/kyverno-version: 1.6.2
    kyverno.io/kubernetes-version: "1.23"
    policies.kyverno.io/subject: Deployment
    policies.kyverno.io/description: >-
      It is common to have two separate Namespaces such as staging and production
      in order to test and promote app deployments in a controlled manner. In order to ensure
      that level of control, certain guardrails must be present so as to minimize regressions or
      unintended behavior. This policy has a set of three rules to try and provide some sane defaults
      for app promotion across these two environments (Namespaces) called staging and production. First,
      it makes sure that every Deployment in production has a corresponding Deployment in staging. Second,
      that a production Deployment uses same image name as its staging counterpart. Third, that
      a production Deployment uses an older or equal image version as its staging counterpart.
spec:
  validationFailureAction: enforce
  background: true
  rules:
#######################
# Ensures that each Deployment being created in production
# has an existing Deployment already in staging of the same name.
  - name: require-staging-deployment
    match:
      any:
      - resources:
          namespaces:
          - production
          kinds:
          - Deployment
    preconditions:
      any:
      - key: "{{request.operation}}"
        operator: In
        value:
        - CREATE        
        - UPDATE
    context:
    - name: deployment_count
      apiCall:
        urlPath: "/apis/apps/v1/namespaces/staging/deployments"
        jmesPath: "items[?metadata.name=='{{ request.object.metadata.name }}'] || `[]` | length(@)"
    validate:
      message: "Every Deployment in production requires a corresponding Deployment in staging."
      deny:
        conditions:
          any:
          - key: "{{deployment_count}}"
            operator: Equals
            value: 0
#######################
# Ensures that each corresponding Deployment in staging and production
# Namespaces uses the same image name (not tag).
  - name: require-same-image
    match:
      any:
      - resources:
          namespaces:
          - production
          kinds:
          - Deployment
    preconditions:
      all:
      - key: "{{request.operation}}"
        operator: In
        value:
        - CREATE        
        - UPDATE
      - key: "{{ deployment_count }}"
        operator: GreaterThan
        value: 0
    context:
    - name: deployment_count
      apiCall:
        urlPath: "/apis/apps/v1/namespaces/staging/deployments"
        jmesPath: "items[?metadata.name=='{{ request.object.metadata.name}}']  || `[]` | length(@)"
    - name: deployment_images
      apiCall:
        urlPath: "/apis/apps/v1/namespaces/staging/deployments/{{ request.object.metadata.name }}"
        jmesPath: "spec.template.spec.containers[].image.split(@, ':')[0]"
    validate:
      message: "Every Deployment in production is required to use the same image(s) as in staging."
      deny:
        conditions:
          any:
          - key: "{{ request.object.spec.template.spec.containers[].image.split(@, ':')[0]  }}"
            operator: AnyNotIn
            value: "{{ deployment_images }}"
#######################
# Ensures that each image version in production is less than or
# equal to its corresponding image version in staging. Uses the container
# name as the basis for comparison. Recommended to pair this policy
# with another policy or rule which enforces only semver image tags, or otherwise extend this rule
# with a precondition which filters out everything else.
  - name: require-same-or-older-imageversion
    match:
      any:
      - resources:
          namespaces:
          - production
          kinds:
          - Deployment
    preconditions:
      all:
      - key: "{{request.operation}}"
        operator: In
        value:
        - CREATE        
        - UPDATE
      - key: "{{ deployment_count }}"
        operator: GreaterThan
        value: 0
    context:
    - name: deployment_count
      apiCall:
        urlPath: "/apis/apps/v1/namespaces/staging/deployments"
        jmesPath: "items[? metadata.name == '{{ request.object.metadata.name }}'  ]  | length(@)"
    - name: deployment_containers
      apiCall:
        urlPath: "/apis/apps/v1/namespaces/staging/deployments/{{ request.object.metadata.name }}"
        jmesPath: "spec.template.spec.containers[]"
    validate:
      message: "Every Deployment in production is required to use an image version less than or equal to the one in staging."
      foreach:
      - list: "request.object.spec.template.spec.containers[]"
        deny:
          conditions:
            any:
            # Uses the container name as the basis for comparing the image tag. Only semver has been tested.
            - key: "{{ element.image.split(@,':')[1] }}"
              operator: GreaterThan
              value: "{{ deployment_containers[?name == '{{element.name}}'].image.split(@,':')[1] | [0] }}"
```

PR #297, opened 2022-04-18
Fix build issue

vyankyGH:
## Related Issue(s)
closes https://github.com/kyverno/kyverno/issues/3604

## Checklist

### DIFF
--- a/cert-manager/limit-duration/kyverno-test.yaml
+++ b/cert-manager/limit-duration/kyverno-test.yaml
@@ -8,7 +8,7 @@ results:
     rule: certificate-duration-max-100days 
     resource: letsencrypt-crt
     kind: Certificate
-    result: skip
+    result: fail
   - policy: cert-manager-limit-duration
     rule: certificate-duration-max-100days 
     resource: acme-crt-short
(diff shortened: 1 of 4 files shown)


chipzoller:
@vyankyGH changing these test to `Fail` is not the right approach as they were previously working with `Skip` which is the intended behavior. I think what we need here is to add [non-existence checks](https://kyverno.io/docs/writing-policies/jmespath/#non-existence-checks). Can you do that instead and see if it works without modifying the tests?

vyankyGH:
> @vyankyGH changing these test to `Fail` is not the right approach as they were previously working with `Skip` which is the intended behavior. I think what we need here is to add [non-existence checks](https://kyverno.io/docs/writing-policies/jmespath/#non-existence-checks). Can you do that instead and see if it works without modifying the tests?

Yup. Will do that

vyankyGH:
@chipzoller I don't think this will get fixed by adding non-existence checks. The response result,  test command getting is _Fail_  instead of skip for all above test cases. this is the reason those test cases getting fail. 
Need to debug more why response changed from Skip to Fail.

chipzoller:
Ok, if it's still failing then we have a CLI code regression somewhere that needs to be identified and fixed.

chipzoller:
Is it ok to close this so the CLI code can be fixed rather than adjusting tests?

vyankyGH:
> Is it ok to close this so the CLI code can be fixed rather than adjusting tests?

Okay. Closing PR.

PR #299, opened 2022-04-22
Added team namespace name validation policy

mvazquezc:
## Description

Adds a policy that validates the name of namespaces/projects created by all authenticated users and only allows namespaces names that start with one of the user's groups followed by a hyphen and something else.

This policy is also great to see how you can work with different JMESpath filters.

I added tests for the policy but I'm not completely sure those make sense since this policy is kind of dynamic in terms of validation (different for each user depending on their groups).

## Checklist


chipzoller (approved):

samj1912:
@chipzoller looks like this policy is causing failures in kyverno CLI tests.

chipzoller:
I have this fixed in a PR.

PR #308, opened 2022-04-26
Update policy

vyankyGH:
## Related Issue(s)
closes https://github.com/kyverno/kyverno/issues/3604

@chipzoller I have fix the failing test from main branch. but for current policy we need to remove the subject part from policy.
Currently we CLI not supporting  _subject_ in match resource.
I'm not sure how it working previously. 
But if we add subject in match resource the rule get skipped as kind does't match kind. 
So I have removed _subject_ from one match resources and now test working as expected.

## Checklist

### DIFF
--- a/best-practices/nginx_path_checks/restrict-path.yaml
+++ b/best-practices/nginx_path_checks/restrict-path.yaml
@@ -33,6 +33,12 @@ spec:
             - key: "{{ request.object.spec.rules[].http.paths[].path.contains(@,'/var/run/secrets') }}"
               operator: AnyIn
               value: [true]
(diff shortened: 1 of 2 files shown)


chipzoller:
This is supposed to be supported in https://github.com/kyverno/kyverno/pull/3188
Also, look at the CI run. It's failing still on an Ingress policy.

vyankyGH:
> This is supposed to be supported in [kyverno/kyverno#3188](https://github.com/kyverno/kyverno/pull/3188) Also, look at the CI run. It's failing still on an Ingress policy.

Correct @chipzoller. Let me check why it's still failing.

JimBugwadia (approved):

Issue #312, opened 2022-04-29
Check deprecated APIs


smcaine:
This ClusterPolicy unfortunately breaks if any of the api's are deprecated, so for example this policy wont get applied to v1.22.x k8s due to following error:\

```
err: kind 'ValidatingWebhookConfiguration' not found in apiVersion 'admissionregistration.k8s.io/v1beta1
```

Would be good to have another suggested way of doing this in the docs.. like what was posted [here](https://github.com/kyverno/kyverno/issues/3261#issuecomment-1068822279) to avoid apply failures, something like:

```yaml
rules:
  - name: validate-v1-22-removals
    match:
      resources:
        kinds:
        - "*"
    preconditions:
      all:
      - key: "{{request.object.apiVersion}}"
        operator: In
        value:
        - admissionregistration.k8s.io/v1beta1
        - apiextensions.k8s.io/v1beta1
        - apiregistration.k8s.io/v1beta1
        - authentication.k8s.io/v1beta1
```

As we use Kustomization to apply a load of other policies, this causes the kustomize to fail and no other policies then get applied.

chipzoller:
Thanks for bringing this up. I meant to update that policy. Can you confirm that the approach listed [here](https://github.com/kyverno/kyverno/issues/3261#issuecomment-1074820180) does work for you?

smcaine:
hey, thanks for quick reply, sorry i did not get chance to try yet..

chipzoller:
If you can check and help test, I'll add this to my list of policies to update for the upcoming 1.7.0 release.

smcaine:
sure thing, i will be able to test this on Tuesday and get back to you, cheers!

chipzoller:
Thank you, and if you can also test for the 1.25 deprecations that would be great.

smcaine:
Hey, yes this all works, have tested both rules successfully.. this only works if you set the `background: true` as wildcards is not supported in background mode.

chipzoller:
Closed in #317

PR #317, opened 2022-05-03
update check-deprecated-apis policy to use wildcard kind + preconditi…

smcaine:
## Related Issue(s)

[https://github.com/kyverno/policies/issues/312](https://github.com/kyverno/policies/issues/312)

## Description

This pr is intended to make the `check-deprecated-apis` policy more user friendly/passive by allowing it to work on a k8s versions that no longer has the api's. For example this ClusterPolicy would fail to apply on k8s > v1.22.x as the apis no longer exist which then causes other issues and other policies not being applied.

## Checklist


chipzoller (approved):
Thank you!

PR #326, opened 2022-05-14
initial check-in of K10 Data Protection Guardrails

leizerbeam:
## Description
Check README.md for more background about these policies for Kasten K10 by Veeam. The guardrails are designed to prevent vulnerabilities associated with misconfiguring data protection backup policies and associated resources (like backup targets). 

## Checklist


kasten/k10-minimum-retention/k10-override-minimum-retentions.yaml
```diff
@@ -0,0 +1,43 @@
...
+spec:
+  rules:
+  - name: k10-policy-minimum-retention
+    match:
+      any:
+      - resources:
+          kinds:
+          - config.kio.kasten.io/v1alpha1/Policy
+    mutate: 
+      # HIPAA example (replace existing values only: ie yearly with 6)
```
chipzoller, line 30: Comments are nice to show users an alternate policy configuration for different use cases, but you may want to give this in a JSON patch form rather than a strategic patch form, which as you know the Kasten Policy resource does not support.

chipzoller (changes requested):
Thanks for this PR! A couple of comments and requests:
* Take a look at the [contribution section of the README](https://github.com/kyverno/policies/blob/main/README.md#contribution) in this repo for some important annotations that should be added. We would want to add `kyverno.io/kyverno-version` and `kyverno.io/kubernetes-version` at a minimum.
* In the `k10-data-protection-by-label` folder there are two Kyverno ClusterPolicy files; one is a `validate` rule and the other is a `generate`. Would you break out the `generate` into its own folder (under the main `kasten` folder)? This is so we can (later) add test cases for it which the Kyverno CLI now supports.
* In all of these Kyverno policies, to be forward compatible, in your `match` statements can you add an `any` block underneath them? Kyverno is moving away from specifying `resources` directly under `match`. Do an `kubectl explain clusterpolicy.spec.rules.match` for more info.

kasten/k10-minimum-retention/patched.yaml
```diff
@@ -0,0 +1,40 @@
+kind: Policy
+apiVersion: config.kio.kasten.io/v1alpha1
+metadata:
+  name: hourly-policy
+  namespace: kasten-io
+    appPriority: Mission-Critical
```
chipzoller, line 6: This line is causing tests to fail because it's invalid. It looks like a label and so should either be placed under `metadata.labels` or simply removed.

kasten/k10-minimum-retention/patched.yaml
```diff
@@ -0,0 +1,41 @@
...
+      - 7
+    weekdays:
+      - 5
+    days:
+      - 15
+  retention:
+    hourly: 24
+    daily: 30
+    weekly: 4
+    monthly: 6
```
chipzoller, line 25: In a previous commit, you had changed `monthly` to `3` except you did not update the patched resource accordingly. As a result, this test is now failing. Would you address, please?

leizerbeam:
hi @chipzoller I am not sure how to fix the DCO issue, I tried to do what the rebase suggested, but the output doesn't really match what I see on my CLI. I tried recloning kyverno/policies then checking out the PR manually but now it does not recognize and changes to commit.

chipzoller (approved):

chipzoller (approved):
Thank you @leizerbeam !

PR #339, opened 2022-06-06
Cleanup values file if only containing  variable `request.operation=CREATE`  and remove variable `request.operation=CREATE`  from values.yaml file

### LINKED ISSUE #338: [Clean-up] In `main` remove variables and references thereto for `request.operation` equals `CREATE`
As of the 1.7.0 CLI, if `request.operation` is not specified in a values file, it will be defaulted to `CREATE`. This results in reduced complexity when writing test cases as there is no longer a need for a values file which defines on the `request.operation` variable. Clean these up.

viveksahu26:
## Related Issue(s)
Closes #338

## Description
From kyverno 1.7 version, the value of `request.operation` is implicit set as `CREATE`. Therefore no need to provide explicit values for `request.operation=CREATE`
--> Cleaned up all values yaml file if only containing one variable, i.e. `request.operation=CREATE` 
--> Also removed variable  `request.operation=CREATE` , from values yaml file

## What does this PR do?
 /cleanup

## Checklist


chipzoller:
Please use either closing keywords or manually link your PR to the issue. This is so merging of the PR automatically closes the issue and creates a link from issue back to PR. I have done so for you this time.

chipzoller (approved):
Thanks, Vivek!

PR #390, opened 2022-09-09
KubeVirt: Introduce rule to enforce instanceType usage

fabiand:
Signed-off-by: Fabian Deutsch <fabiand@fedoraproject.org>

## Related Issue(s)

None

## Description

Introduces a basic rule in order to enforce the usage of instanceTypes and preferences, and prevents the usage of domain templates in VMs.

## Checklist

### DIFF
--- a/kubevirt/k6t-only-instancetype/k6t-require-instancetype.yaml
+++ b/kubevirt/k6t-only-instancetype/k6t-require-instancetype.yaml
@@ -1,0 +1,30 @@
+apiVersion: kyverno.io/v1
+kind: ClusterPolicy
+metadata:
+  name: k6t-require-instancetype
+  annotations:
+    policies.kyverno.io/title: Require instanceType
+    policies.kyverno.io/category: KubeVirt
+    kyverno.io/kyverno-version: 1.6.2
+    policies.kyverno.io/minversion: 1.6.0
+    kyverno.io/kubernetes-version: "1.25"
+    policies.kyverno.io/subject: VirtualMachine
+    policies.kyverno.io/description: >-
+      Check VirtualMachines and validate that they are not using a domain template (`.spec.template.domain`)
+      but instead are based on instaceTypes and preferences only.
+spec:
+  validationFailureAction: enforce
+  rules:
+  - name: k6t-dont-allow-domain-template
+    match:
+      any: 
+      - resources:
+          kinds:
+          - VirtualMachine
+    validate:
+      message: "VirtualMachines must only use instanceTypes and preferences, a domain template is not allowed."
+      pattern:
+        spec:
+          template:
+            spec:
+              domain: null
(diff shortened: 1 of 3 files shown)


fabiand:
@chipzoller :wave: Can you allow me to get this PR tested?

kubevirt/k6t-only-instancetype/k6t-require-instancetype.yaml
```diff
@@ -0,0 +1,29 @@
+apiVersion: kyverno.io/v1
+kind: ClusterPolicy
+metadata:
+  name: k6t-require-instancetype
+  annotations:
+    policies.kyverno.io/title: Require instanceType
+    policies.kyverno.io/category: KubeVirt
+    kyverno.io/kyverno-version: 1.6.2
```
chipzoller, line 8: This annotation means you tested and verified it on 1.6.2. Is that correct?

kubevirt/k6t-only-instancetype/k6t-require-instancetype.yaml
```diff
@@ -0,0 +1,29 @@
...
+    policies.kyverno.io/minversion: 1.6.2
```
chipzoller, line 9: Minversion for this policy should be 1.6.0

kubevirt/k6t-only-instancetype/k6t-require-instancetype.yaml
```diff
@@ -0,0 +1,29 @@
...
+    kyverno.io/kubernetes-version: "1.21-1.25"
```
chipzoller, line 10: This annotation means you personally tested and verified its application on Kubernetes 1.21, 1.22, 1.23, 1.24, and 1.25. Was that the case? If not, please only specify the version(s) on which you actually tested it.

kubevirt/k6t-only-instancetype/k6t-require-instancetype.yaml
```diff
@@ -0,0 +1,29 @@
...
+  validationFailureAction: enforce
+  rules:
+  - name: k6t-dont-allow-domain-template
+    match:
+      any: 
+      - resources:
+          kinds:
+          - VirtualMachine
+    validate:
+      message: "VirtualMachines must only use instanceTypes and preferences, a domain template is not allowed."
```
chipzoller, line 25: Based on your message, it looks like your pattern may be wrong. This pattern effectively means, "make sure a VirtualMachine has `spec.template.domain` with something inside it.

chipzoller (changes requested):
Thanks for the contribution! Tests are failing because you appear to have swapped the VirtualMachine resources (pass should be fail and vice versa).

kubevirt/k6t-only-instancetype/k6t-require-instancetype.yaml
fabiand: *cogh* I was optimistic, I'm still trying to run minikube on Silverblue in order to verify it.

kubevirt/k6t-only-instancetype/k6t-require-instancetype.yaml
fabiand: Thanks! I also noticed that I'm checking the wrong path. I'll fix both.

chipzoller (approved):

PR #403, opened 2022-09-27
Add testcases for PSS policies

vyankyGH:
## Related Issue(s)
Pss policy test cases

## Checklist


chipzoller (changes requested):
Please write all policies with any or all blocks.

vyankyGH:
> Please write all policies with any or all blocks.

Correct, How can I miss that 😔. will update

pss/privileged-containers/policy.yaml
```diff
@@ -0,0 +1,21 @@
...
+          kinds:
+          - Pod
+    validate:
+      podSecurity:
+        level: baseline
+        version: v1.24
+        exclude:
+        - controlName: Privileged Containers
+          images:
+          - nginx*
```
chipzoller, line 21: You're writing the wildcard capability but you aren't testing for it. Please add a test to explicitly check this code path.

chipzoller (changes requested):
Can we also test for just control name exclusion without images? I assume this capability is still present after Shuting's PR.

Also, any of these controls which have checks at the Pod spec and container level must test for both.

vyankyGH:
> Can we also test for just control name exclusion without images? I assume this capability is still present after Shuting's PR.
> 
> Also, any of these controls which have checks at the Pod spec and container level must test for both.

okay

pss/privileged-containers/policy.yaml
realshuting: Without `images` the checks are performed at the spec level.

chipzoller: So in this case for Privileged Containers, is defining one or more images required because it has no `spec` fields to enforce?

realshuting: Yes, I'm adding the validation checks in.

chipzoller: Does `images` allow a single wildcard character to match on all image names?

realshuting: Yes, wildcard match is supported.

chipzoller:
Sorry for the delay here. Since I had already added some podSecurity samples prior to merging this, I'll probably just save you some work and take what you've done and format it for the existing directory structure.

chipzoller (approved):

chipzoller:
Superseded by PR #427

PR #405, opened 2022-09-27
feat: add update operation to the list

developer-guy:

## Description

There might be existing TaskRuns that can be modified with anyone else, so we should block updating them too.

/cc @chipzoller 

## Checklist

### DIFF
--- a/tekton/block-tekton-task-runs/block-tekton-task-runs.yaml
+++ b/tekton/block-tekton-task-runs/block-tekton-task-runs.yaml
@@ -30,8 +30,10 @@ spec:
     preconditions:
       all:
       - key: "{{ request.operation }}"
-        operator: Equals
-        value: CREATE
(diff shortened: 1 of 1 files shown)


tekton/block-tekton-task-runs/block-tekton-task-runs.yaml
```diff
@@ -30,8 +30,10 @@ spec:
     preconditions:
       all:
       - key: "{{ request.operation }}"
-        operator: Equals
-        value: CREATE
+        operator: AnyIn
+        value:
```
Dentrax, line 34: should we also add `PATCH`?

chipzoller: No such thing from a webhook standpoint.

chipzoller (approved):

PR #424, opened 2022-10-14
check type only if set

monotek:
Signed-off-by: André Bauer <monotek@users.noreply.github.com>

## Description

* checks "type" only if set otherwise rule will also fail on default, which is "ClusterIP", when the type is not set

## Checklist


chipzoller (approved):
Thanks for the suggested improvement!

PR #436, opened 2022-10-20
run pdb check only if replicas >= 3

monotek:
Signed-off-by: André Bauer <andre.bauer@staffbase.com>

## Description

* run pdb check only if replicas >= 3 as pdb are usualy not defined for single pod deployments
* also add StatefulSet as resource
* use any in rule

## Checklist

```
kyverno test . 

Executing require-pdb...

│───│─────────────│─────────────│──────────────────────────────────────│────────│
│ # │ POLICY      │ RULE        │ RESOURCE                             │ RESULT │
│───│─────────────│─────────────│──────────────────────────────────────│────────│
│ 1 │ require-pdb │ require-pdb │ default/Deployment/nginx-deploy-pass │ Pass   │
│ 2 │ require-pdb │ require-pdb │ default/Deployment/nginx-deploy-skip │ Pass   │
│───│─────────────│─────────────│──────────────────────────────────────│────────│

Test Summary: 2 tests passed and 0 tests failed


```

### DIFF
--- a/other/require_pdb/require_pdb.yaml
+++ b/other/require_pdb/require_pdb.yaml
@@ -17,26 +17,31 @@ spec:
   validationFailureAction: audit
   background: false
   rules:
-  - name: require-pdb
-    match:
-      resources:
-        kinds:
-        - Deployment
-    preconditions:
-      any:
-      - key: "{{request.operation || 'BACKGROUND'}}"
-        operator: Equals
-        value: CREATE
-    context:
-    - name: pdb_count
-      apiCall:
-        urlPath: "/apis/policy/v1/namespaces/{{request.namespace}}/poddisruptionbudgets"
-        jmesPath: "items[?label_match(spec.selector.matchLabels, `{{request.object.spec.template.metadata.labels}}`)] | length(@)"
-    validate:
-      message: "There is no corresponding PodDisruptionBudget found for this Deployment."
-      deny:
-        conditions:
-          any:
-          - key: "{{pdb_count}}"
-            operator: LessThan
-            value: 1
+    - name: require-pdb
+      match:
+        any:
+          - resources:
+              kinds:
+                - Deployment
+                - StatefulSet
+      preconditions:
+        all:
+          - key: "{{request.operation || 'BACKGROUND'}}"
+            operator: Equals
+            value: CREATE
+          - key: "{{ request.object.spec.replicas }}"
+            operator: GreaterThanOrEquals
+            value: 3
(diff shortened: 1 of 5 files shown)


other/require_pdb/require_pdb.yaml
```diff
@@ -23,10 +23,13 @@ spec:
...
         - Deployment
     preconditions:
-      any:
+      all:
       - key: "{{request.operation || 'BACKGROUND'}}"
         operator: Equals
         value: CREATE
+      - key: "{{ request.object.spec.replicas }}"
+        operator: GreaterThanOrEquals
+        value: "3"
```
chipzoller, line 32: You want to leave this unquoted so it's an integer.

chipzoller (changes requested):
Minor thing. And as an "extra", it would be great if you could add maybe a couple more tests, including a negative test.

monotek:
update the tests so one example with "skip" is available too.

other/require_pdb/require_pdb.yaml
```diff
@@ -17,26 +17,29 @@ spec:
...
-      message: "There is no corresponding PodDisruptionBudget found for this Deployment."
-      deny:
-        conditions:
-          any:
-          - key: "{{pdb_count}}"
-            operator: LessThan
-            value: 1
\ No newline at end of file
+    - name: require-pdb
+      match:
```
chipzoller, line 21: Since you're updating this, go ahead and add `any`/`all` under the `match` object here to bring it current.

other/require_pdb/require_pdb.yaml
```diff
@@ -17,26 +17,29 @@ spec:
...
+          kinds:
+            - Deployment
+      preconditions:
+        all:
+          - key: "{{request.operation || 'BACKGROUND'}}"
+            operator: Equals
+            value: CREATE
+          - key: "{{ request.object.spec.replicas }}"
+            operator: GreaterThanOrEquals
+            value: "3"
```
chipzoller, line 32: Don't need quotes around integers.

chipzoller (changes requested):

other/require_pdb/require_pdb.yaml
monotek: done

other/require_pdb/require_pdb.yaml
monotek: done

chipzoller (approved):

PR #468, opened 2022-12-10
fix allnotin policy

cleverhu:
## Description
fix allnotin policy.

## Checklist


chipzoller (commented):
I'm not clear why changes are needed here.

cleverhu:
This is to solve the problem of the pr(https://github.com/kyverno/kyverno/pull/5636) pipeline.Because there was a problem with the previous method, I fixed it.

cleverhu:
The previous logic of the method was that all keys must be included in value the it will return false. In fact, as long as any key is included in value, it should return false.

chipzoller (approved):

Issue #509, opened 2023-02-27
Prepend Image Registry: initcontainers rule uses name instead of path for the image path


michael-barker:
In the [prepend-registry policy](https://kyverno.io/policies/other/prepend_image_registry/prepend_image_registry/), the `prepend-registry-initcontainers` rule has a bug where the name is of the image is specified where this should be the path.

Currently the image is set to the follow.
```
image: registry.io/{{ images.initContainers."{{element.name}}".name}}:{{images.initContainers."{{element.name}}".tag}}
```

This should be.
```
image: registry.io/{{ images.initContainers."{{element.name}}".path}}:{{images.initContainers."{{element.name}}".tag}}
```

chipzoller:
Thank you for the report. Would you be interested in submitting a PR for this minor fix?

michael-barker:
@chipzoller Working on one now :grin:

michael-barker:
Another thing that I've noticed with this policy, if you submit an image that uses a digest then the image format becomes invalid. I haven't worked with Kyverno enough to be able to implement.

PR #510, opened 2023-02-27
Fix image path for prepend_image_registry policy

michael-barker:
## Related Issue(s)

Resolves: #509

## Description

Fix the image path for the prepend-registry-initcontainers rule so that it uses the image path for the path instead of the image name. We also update the resources used in the tests to use an image containing a path.

## Checklist


chipzoller (approved):

chipzoller:
Please sign-off on your commits (DCO)

PR #526, opened 2023-03-20
Added example policy to automate K10 policy creation based on SLA pre…

mattbator:
## Related Issue(s)

N/A

## Description

New example policy to enforce setting a dataprotection label on a newly created namespace, corresponding to a K10 policy preset, which will then be used to generate a K10 backup policy based on the centrally configured preset settings.

Generate policy placed in the same directory as the two policies are designed to be used together.

## Checklist


chipzoller:
Please sign off on your commits (DCO).

PR #528, opened 2023-03-22
restrict_node_selection - reduce spec.nodename false-positives

### LINKED ISSUE #512: Restrict node selection
Hi everyone,
can someone give me some context about the usage of the policy regarding `spec.nodeName`. https://kyverno.io/policies/other/restrict_node_selection/restrict_node_selection
How can you prevent this warning?  
If you don't set nodeSelector and nodeName in a deployment, the resulting pod will have a set spec.nodeName  
```
12s Warning PolicyViolation pod/redis-server-master-0 policy restrict-node-selection/restrict-nodename fail: validation error: Setting the nodeName field is prohibited. rule restrict-nodename failed at path /spec/nodeName/
```

### LINKED ISSUE #525: Restrict node selection: every Pod gets a .spec.nodeName, every Pod triggers a policy error
This field exists on every Pod once they are scheduled. This Policy will create audit warnings 100% of the time.

### LINKED ISSUE #529: Restrict node selection
apiVersion: v1
kind: Pod
metadata:
  name: ubuntu
  namespace: ubuntu
  labels:
    app: ubuntu
spec:
  securityContext:
    seccompProfile:
      type: RuntimeDefault
  containers:
  - image: ubuntu:20.04
    command:
      - "sleep"
      - "604800"
    imagePullPolicy: IfNotPresent
    name: ubuntu
    resources:
      limits:
        memory: "6400Mi"
      requests:
        cpu: "500m"
        memory: "6400Mi"
    securityContext:
      capabilities:
        drop:
        - ALL
        - CAP_NET_RAW
      allowPrivilegeEscalation: false
      runAsNonRoot: true
      readOnlyRootFilesystem: true
      runAsUser: 1001
  nodeName: null
  restartPolicy: Always

fadecore:
The current behaviour is scanning all running pods for the definition of spec.nodename.  
This will always be true, because after the deployment of a pod it is filled with the worker nodename the pod is running at.  
As a result we have many false-positive events.

## Related Issue(s)
closes #525 
closes #512 
Closes #529

## Description

What we really want to make sure is, that a user is not targeting a specific nodename in it's deployment files at creation time.
This way it is up to the scheduler to choose the best node, especially during rolling updates or renaming of worker nodes.

With adding a precondition to check only on CREATE events we make sure deployed pods are not triggering events anymore.

## Checklist

### DIFF
--- a/other/restrict_node_selection/restrict_node_selection.yaml
+++ b/other/restrict_node_selection/restrict_node_selection.yaml
@@ -37,9 +37,8 @@ spec:
     preconditions:
       all:
       - key: "{{request.operation || 'BACKGROUND'}}"
-        operator: Equal
-        value:
-        - CREATE
+        operator: Equals
+        value: CREATE
     validate:
       message: Setting the nodeName field is prohibited.
       pattern:
--- a/other/restrict_node_selection/restrict_node_selection.yaml
+++ b/other/restrict_node_selection/restrict_node_selection.yaml
@@ -34,6 +34,12 @@ spec:
       - resources:
           kinds:
           - Pod
+    preconditions:
+      all:
+      - key: "{{request.operation || 'BACKGROUND'}}"
+        operator: Equal
+        value:
+        - CREATE
     validate:
       message: Setting the nodeName field is prohibited.
       pattern:


other/restrict_node_selection/restrict_node_selection.yaml
```diff
@@ -34,8 +34,14 @@ spec:
       - resources:
           kinds:
           - Pod
+    preconditions:
+      all:
+      - key: "{{request.operation || 'BACKGROUND'}}"
+        operator: Equal
```
chipzoller, line 40: Operator name is `Equals`.

remram44:
Does this show a user to say the nodeName if they create without it and then update? Or is the Pod immutable right away?

other/restrict_node_selection/restrict_node_selection.yaml
```diff
@@ -34,8 +34,14 @@ spec:
...
+        value:
```
chipzoller, line 41: Value of `value` should be stated as a string; no need for a list.

chipzoller (changes requested):

chipzoller:
> Does this allow a user to set the nodeName if they create without it and then update? Or is the Pod immutable right away?

Setting the nodeName after the Pod is created has no effect.

other/restrict_node_selection/restrict_node_selection.yaml
fadecore: Thanks! Hint for me:
Use official docs https://kyverno.io/docs/writing-policies/preconditions and not an existing ClusterPolicy.

fadecore (commented):
thx for hints

remram44:
> Setting the nodeName after the Pod is created has no effect.

What if you put a nodeSelector or resource requirements that can't be met, so the Pod stays Pending? Then you update it to remove that and add nodeName? Would this allow you to do what the Policy is meant to prevent?

chipzoller:
> > Setting the nodeName after the Pod is created has no effect.
> 
> What if you put a nodeSelector or resource requirements that can't be met, so the Pod stays Pending? Then you update it to remove that and add nodeName? Would this allow you to do what the Policy is meant to prevent?

I don't know what the detailed behavior is here. Would have to test and see how the kube-scheduler responds.

chipzoller (approved):

remram44:
The rule is:

> pod updates may not change fields other than `spec.containers[*].image`, `spec.initContainers[*].image`, `spec.activeDeadlineSeconds`, `spec.tolerations` (only additions to existing tolerations) or `spec.terminationGracePeriodSeconds` (allow it to be set to 1 if it was previously negative)

So I don't think there is a way around this policy for pods after all. Thanks for the fix!

chipzoller:
@fadecore with this change, we also need to change `background: false` or else Policy Reports will always have failed entries. Is this a change you'd also like to make?

chipzoller:
#533

PR #592, opened 2023-04-29
Minor fixes for JSON patch policies

chipzoller:
## Description

Fixes and improves some JSON patch policies.

## Checklist

### DIFF
--- a/kasten/k10-minimum-retention/artifacthub-pkg.yml
+++ b/kasten/k10-minimum-retention/artifacthub-pkg.yml
@@ -19,4 +19,4 @@ annotations:
   kyverno/category: "Kasten K10 by Veeam"
   kyverno/kubernetesVersion: "1.21-1.22"
   kyverno/subject: "Policy"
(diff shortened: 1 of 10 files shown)


eddycharly:
@chipzoller why do we need to disable autogen ?

chipzoller:
It was due to #4731.

eddycharly:
But this is fixed now right ?

chipzoller:
It's fixed in main but not in alpha.2, which is what I was testing against. I thought it wouldn't be bad to leave it enabled in these policies for now.

eddycharly (approved):

eddycharly:
Ok but I really want to fix that, we should be able to support json patches but we need to revisit our autogen approach.

chipzoller:
Ok, so what do you want me to do here? Re-enable autogen?

eddycharly:
Nope, I approved. We will want to enable again when we support autogen with json patches.

Issue #581, opened 2023-04-20
[Test] change digest value for all the artifact-pkg.yml in the repo


guohaoyu110:
```
#!/bin/bash

# Loop through every directory containing a .yaml file
for dir in $(find . -name "*.yaml" -exec dirname {} \; | sort -u); do
    # Get the directory name without the path
    dirname=$(basename $dir)

    # Check if a file with the same name as the directory exists in that directory
    if [ -f "$dir/$dirname.yaml" ]; then
        # Calculate the SHA256 hash of the file
        sha256=$(shasum -a 256 "$dir/$dirname.yaml" | cut -d' ' -f1)

        # Add the hash to the artifacthub-pkg.yml file in the same directory in the format "digest: <hash>"
        echo "digest: $sha256" >> "$dir/artifacthub-pkg.yml"
    fi
done

```

PR #597, opened 2023-05-02
Update digest field for every artifacthub-pkg.yml

guohaoyu110:
## Related Issue(s)
Close: https://github.com/kyverno/policies/issues/581

## Description
Update digest field for every artifacthub-pkg.yml

## Checklist

### DIFF
--- a/README.md
+++ b/README.md
@@ -84,7 +84,6 @@ description: >-
       automatically annotates Pods (and Pod controllers) which refer to a PVC so that
       all volumes are listed in the aforementioned annotation if a Namespace with the label
       `velero-backup-pvc=true`.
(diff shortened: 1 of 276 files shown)


guohaoyu110:
#!/bin/bash

# Loop through every directory containing a .yaml file
for dir in $(find . -name "*.yaml" -exec dirname {} \; | sort -u); do
    # Get the directory name without the path
    dirname=$(basename $dir)

    # Check if a file with the same name as the directory exists in that directory
    if [ -f "$dir/$dirname.yaml" ]; then
        # Calculate the SHA256 hash of the file
        sha256=$(shasum -a 256 "$dir/$dirname.yaml" | cut -d' ' -f1)

        # Add the hash to the artifacthub-pkg.yml file in the same directory in the format "digest: <hash>"
        echo "digest: $sha256" >> "$dir/artifacthub-pkg.yml"
    fi
done

guohaoyu110:
@chipzoller

chipzoller (approved):

PR #612, opened 2023-05-15
moves the validation step above the first test

guohaoyu110:
## Related Issue(s)
Close: https://github.com/kyverno/policies/issues/611

## Description
moves the validation step above the first test and delete one unnecessary file 

## Checklist


chipzoller:
Should only be one file being modified here.

guohaoyu110:
but if remain the enforce/kustomization.yaml, the check workflow cannot pass because that file does not meet the requirement.

chipzoller:
Ok, I see. I didn't think about Kustomization files. Add an exclusion to the logic. `kustomization.yaml` is exempt since this is not a policy.

guohaoyu110:
> Ok, I see. I didn't think about Kustomization files. Add an exclusion to the logic. `kustomization.yaml` is exempt since this is not a policy.

you mean i add exemption in the ci.yml workfolw to ignore the kustomization files?

chipzoller:
Yes, we don't care about `kustomization.yaml` and it is exempt from these checks.

PR #615, opened 2023-05-17
modify kyverno policies to use preconditions.conditions

MariamFahmy98:
## Description
As a result of supporting cel expressions in preconditions for `validate.cel` subrule, the preconditions are modified to be as follows:
```
preconditions:
   conditions:
     any:
      - key: "{{ request.object.metadata.labels.color || '' }}"
        operator: Equals
        value: blue
```
OR using cel expressions in case of `validate.cel` subrule as follows:
```
preconditions:
  celConditions: 
    - name: "Condition Name"
      expression: "cel expression"
```

This PR is for modifying all policies that make use of `preconditions`.

### DIFF
--- a/argo/application-prevent-default-project/application-prevent-default-project.yaml
+++ b/argo/application-prevent-default-project/application-prevent-default-project.yaml
@@ -23,10 +23,11 @@ spec:
             kinds:
               - Application
       preconditions:
-        all:
-        - key: "{{ request.operation || 'BACKGROUND' }}"
-          operator: NotEquals
-          value: DELETE
+        conditions:
+          all:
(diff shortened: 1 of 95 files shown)


chipzoller:
@MariamFahmy98 we can't do this because it will break backwards compatibility for all these policies in the library.

MariamFahmy98:
> @MariamFahmy98 we can't do this because it will break backwards compatibility for all these policies in the library.

Thanks for the info. I mainly worked on it because of `cli-test` failure in the CEL support [PR](https://github.com/kyverno/kyverno/pull/7070)

PR #622, opened 2023-05-23
Have policies apply on UPDATE when appropriate

remram44:
## Description

Some policies only validate on `CREATE`, but apply to mutable objects. Let's validate on `UPDATE` as well.

I have checked that current policies can be bypassed using `kubectl apply`, `kubectl patch`, or `kubectl edit`, which is a big problem.

## Checklist

# Additional notes

I have not updated the policies that are labelled as examples and look to not be complete solutions. I have also ignored policies that use resource types I am not familiar with, such as `PipelineRun`, so it is possible a similar change needs to be made for the policies of those custom resources.

### DIFF
--- a/other/ingress-host-match-tls/ingress-host-match-tls.yaml
+++ b/other/ingress-host-match-tls/ingress-host-match-tls.yaml
@@ -29,8 +29,10 @@ spec:
     preconditions:
       all:
       - key: "{{request.operation || 'BACKGROUND'}}"
-        operator: Equals
-        value: CREATE
+        operator: AllIn
+        value:
+        - CREATE
+        - UPDATE
     validate:
       message: "The host(s) in spec.rules[].host must match those in spec.tls[].hosts[]."
       deny:
--- a/other/ingress-host-match-tls/ingress-host-match-tls.yaml
+++ b/other/ingress-host-match-tls/ingress-host-match-tls.yaml
@@ -29,7 +29,7 @@ spec:
     preconditions:
       all:
       - key: "{{request.operation || 'BACKGROUND'}}"
-        operator: AllIn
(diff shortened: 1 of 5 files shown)


remram44:
Respectful bump.

I am already using the updated policies, but is there something I should do here? Do you need additional tests? (not sure if that can be done for UPGRADE)

chipzoller (changes requested):
Couple of things. The `AnyIn` operator is more appropriate here even though, yes, functionally `AllIn` performs the same.

Second, when updating policies you must calculate the new sha256 hashes of them and update `artifacthub-pkg.yml` in each directory so that Artifact Hub will see the changes.

remram44:
I have changed to `AnyIn` as requested.

I don't know what artifacthub is or how your hashes are generated, it's not documented anywhere either. If you really expect contributors to do this, you should probably document it, or better yet have GitHub CI do it (if you want contributions, that is).

remram44:
If DCO is required, that should probably also be mentioned in your contribution section...

Look, I have fixed what seems to me like a security issue. I am willing to go through your hoops to have the rest of your users benefit from this, but you seem to expect that all contributors go through many such hoops, none of them listed anywhere, to appease the automated checks that you enforce on us.

I know this is open source and we all have day jobs that take priority etc, but surely you can see that something serious is broken with your process?

What's an artifacthub digest? Do I literally hash the file?

chipzoller:
I'm sorry you've found this process so difficult. I don't recall anyone else has ever had this amount of difficulty with contributing. In any case, I've made these changes myself in PR #636 and also tried to improve the contribution guidelines so this is easier going forward. I will close this PR since all changes are accounted for in #636.

PR #627, opened 2023-06-01
Add add-karpenter-daemonset-priority-class

wdonne:
## Description

When a DaemonSet is added to a cluster every node will get a new pod. There may not be enough room for this on every node. Karpenter cannot provision extra nodes just for the DaemonSet because the new pods are not scheduled the way regular pods are. It would require parallel scheduling logic that is not proper to Kubernetes. Therefore, eviction of regular  pods should happen instead. This can be achieved with the priority class system-node-critical.

## Checklist


chipzoller:
Thank you for this PR. Couple things. Could you merge main and also sign off on your commits, please?

wdonne:
Hi Chip,

I have to admit I never heard of signing off commits before. I will follow the procedure described in DCO to fix it.

The PR was created off the main branch, as requested in the contributions section. Do you mean I should merge main when fixing the DCO?

Best regards,

Werner.

wdonne:
I will start over with a new pull request because I'm afraid of breaking something.

PR #632, opened 2023-06-02
test: aws, best-practices

mintbomb27:
## Description

Kuttl Tests for 
* `require-encryption-aws-loadbalancers`
* `add-networkpolicy`
* `add-networkpolicy-dns`
* `add-ns-quota`
## Checklist


chipzoller (approved):
These look good.

One comment to suggest in future tests....

For any tests which need a Namespace created as a step in the test case, we like to try and name those Namespaces somewhat uniquely. This is for a couple reasons:

1. It helps us find logs quicker if only one test cases uses that Namespace
2. a previous test which might use the same Namespace, if there are problems removing it, does not impact subsequent tests

PR #633, opened 2023-06-03
tests: best-practices tests

mintbomb27:
## Description

Tests for:

- [x] add-safe-to-evict
- [x] restrict-service-external-ips
- [x] restrict-node-port
- [x] restrict-image-registries
- [x] require-pod-requests-limits
- [x] require-probes
- [x] restrict-ro-rootfs
- [x] require-labels
- [x] require-drop-cap-net-raw
- [x] require-drop-all
- [x] disallow-latest-tag
- [x] disallow-helm-tiller
- [x] disallow-empty-ingress-host
- [x] disallow-default-namespace
- [x] disallow-cri-sock-mount

## Checklist


chipzoller:
Is this PR ready for review?

best-practices/disallow-helm-tiller/02-manifests.yaml
chipzoller, line 1: Same situation here as the previous comment. What about if a Pod has two containers where one is "good" and the other is "bad"? The entire Pod must be rejected at that point. You could even forego the simpler version of these tests in favor of the 2-container examples if you wanted. But, as I said, tests like this are critical. Far too often we have _thought_ that code worked but it turns out it didn't because we were only testing situations where a given resource had only one "thing" which was being validated. We will not make that mistake going forward.

best-practices/require-drop-all/02-manifests.yaml
chipzoller, line 1: Ditto again. You can borrow test resources I wrote here: https://github.com/kyverno/policies/blob/main/pod-security/restricted/disallow-capabilities-strict/resource.yaml

best-practices/require-drop-cap-net-raw/02-manifests.yaml
chipzoller, line 1: Ditto to above.

best-practices/require-labels/02-manifests.yaml
chipzoller, line 1: Since auto-gen is going to be active for this one, we should ensure at least one resource from each class (Pod, Deployment, CronJob) is also being tested.

best-practices/require-pod-requests-limits/02-manifests.yaml
chipzoller, line 1: Similar comment as those above. What if a Pod has two containers where one is fully compliant and the other is not?

best-practices/require-probes/02-manifests.yaml
chipzoller, line 1: Also, auto-gen. And also, it would be good to test the UPDATE side of this policy: What happens when there's a compliant Pod (which passed the policy) and then it is subsequently updated to add a new container which is not compliant. Is the update blocked?

best-practices/require-ro-rootfs/02-manifests.yaml
chipzoller, line 1: Auto-gen

best-practices/restrict-image-registries/02-manifests.yaml
chipzoller, line 1: On this test, let's also test for the ephemeralContainers. If you haven't used this before, we'll be doing a `kubectl debug` along with an in-line Pod definition to add a non-compliant ephemeral container.

best-practices/add-safe-to-evict/02-manifests.yaml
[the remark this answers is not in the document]
mintbomb27: Okay. Adding it.

best-practices/disallow-cri-sock-mount/good-pod.yaml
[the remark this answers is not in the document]
mintbomb27: Done!

best-practices/disallow-helm-tiller/02-manifests.yaml
mintbomb27: Got that.

best-practices/require-drop-all/02-manifests.yaml
mintbomb27: Should I cover the Pod, CronJob, and Deployment here as well?

chipzoller: It's a good idea. Basically, any policy where auto-gen is enabled (even if selectively enabled) we should try and have one test resource per auto-generated rule.

mintbomb27: How do we figure if auto-gen is enabled?

chipzoller: Give [the docs](https://kyverno.io/docs/writing-policies/autogen/) a read for starters, then as you develop these tests and create the policies, get the policy back and inspect the YAML. You should see, if auto-gen is active, additional rules created under the `status` object.

best-practices/require-probes/02-manifests.yaml
mintbomb27: For some reason, this doesn't auto-gen a rule for a Job, or CronJob.

chipzoller: Yes, because it has the auto-gen override annotation set: `pod-policies.kyverno.io/autogen-controllers: DaemonSet,Deployment,StatefulSet`

mintbomb27: Woah. Just saw that.

best-practices/require-probes/bad-pod-update.yaml
chipzoller, line 1: This update would be invalid. A container name may not be changed and two containers in the same Pod cannot have the same name. Your update test here should only try and add a second container (with a different name) to the existing Pod but with no probes defined.

best-practices/require-probes/good-pods.yaml
```diff
@@ -0,0 +1,56 @@
...
+      grpc:
+        port: 8888
+    readinessProbe:
+      tcpSocket:
+        port: 8080
+      periodSeconds: 10
+    livenessProbe:
+      tcpSocket:
+        port: 7070
+      periodSeconds: 20
```
chipzoller, line 21: I would also just choose one probe here and not all 3 as it's unrealistic, but nevertheless still works.

chipzoller (changes requested):
Looks great! Only a small nit regarding the UPDATE test.

PR #635, opened 2023-06-07
tests: istio and karpenter tests

mintbomb27:
## Description

Covers 7 tests from istio, and all from karpenter.

## Checklist

### DIFF
--- a/istio/create-authorizationpolicy/02-clusterrole.yaml
+++ b/istio/create-authorizationpolicy/02-clusterrole.yaml
@@ -1,0 +1,17 @@
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRole
+metadata:
+  name: kyverno:background-controller:additional
+  labels:
+    app.kubernetes.io/component: background-controller
+    app.kubernetes.io/instance: kyverno
+    app.kubernetes.io/part-of: kyverno
+rules:
+- apiGroups:
+  - security.istio.io
+  resources:
+  - authorizationpolicies
+  verbs:
+  - create
+  - update
+  - delete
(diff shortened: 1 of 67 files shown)


istio/create-authorizationpolicy/02-clusterrole.yaml
```diff
@@ -0,0 +1,17 @@
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRole
+metadata:
+  name: kyverno:background-controller:additional
```
chipzoller, line 4: Here and in subsequent tests where we need to add custom permissions, don't use the "additional" suffix. Make it meaningful for each test. For example, here you could the suffix "istio-authzpolicy" or something. The more specific and less generic we make any of the resources we create, the easier it is to track down issues in the kuttl output log or the Pod logs. If multiple tests are using the same generic names, if we have to go searching through logs it can be much more time consuming and difficult.

chipzoller (changes requested):
Look very good. One very minor nit.

istio/create-authorizationpolicy/02-clusterrole.yaml
mintbomb27: Makes sense. I'll change it.

istio/require-authorizationpolicy/03-cpolr.yaml
chipzoller, line 1: You're going to have to wait for probably 30 seconds before trying this.

mintbomb27: I'll wait and then assert? Or just assert?

chipzoller (approved):
Grand

PR #644, opened 2023-06-13
kubevirt: Only enforce use of instance types and preferences

lyarwood:
## Description

The policy would previously check for the use of `spec.template.domain` that is not actually invalid when using instance types and preferences.

This change drops that check and replaces it with two validations ensuring both `spec.instancetype.name` and `spec.preference.name` are provided. Any remaining validation being left to the submission webhooks within KubeVirt itself.

## Checklist

### DIFF
--- a/kubevirt/README.md
+++ b/kubevirt/README.md
@@ -6,4 +6,4 @@ This directory contains a few examples in order to apply policies on
 Examples
 
 - add-services - Add a service for every VMI which is getting created. This is an example for a generating policy.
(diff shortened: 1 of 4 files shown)


chipzoller (approved):
LGTM, thanks!

chipzoller:
We need to actually fix the test here.

@mintbomb27 are you able to push to this PR and fix the test?

fabiand:
LGTM, many thanks Lee

chipzoller:
Will be followed up with a patch. Thanks again.

PR #648, opened 2023-06-14
feat: kyverno-kuttl from main

### LINKED ISSUE #647: Test with Kyverno's kuttl tool from `main`
[Here](https://github.com/kyverno/policies/blob/96c8d13b9e4374fa4be68bf57f0f0b37ac3efb6a/.github/workflows/test.yml#L63) we are pinning to a specific versions whereas we should do this for `main` branch of Kyverno's fork of kuttl. While this generally isn't a good practice, in our case we control the code and infrequently update it. Having new changes automatically picked up rather than having to change multiple workflows which point to it will be helpful.

@mintbomb27 can you make this minor change, please?

mintbomb27:
## Related Issue(s)
Fixes #647 

## Description
Gets the kyverno kuttl tool from main, instead of a predefined version.

## Checklist

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -60,7 +60,7 @@ jobs:
       - name: Install Tools
         run: |
           GOBIN=$(pwd)/.tools/ go install sigs.k8s.io/kind@v0.19.0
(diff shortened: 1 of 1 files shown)


chipzoller (approved):
Thanks, Alok

Issue #546, opened 2023-03-30
Kuttl tests for all policies


chipzoller:
Kyverno conducts its end-to-end testing using a tool called [kuttl](https://kuttl.dev/) which has [been forked](https://github.com/kyverno/kuttl) and significantly enhanced with features designed to test concerns that policy engines have. While the Kyverno project has a [significant library of tests](https://github.com/kyverno/kyverno/tree/main/test/conformance/kuttl) which cover various facets of Kyverno, there are no tests which cover each of the policies in the policy library here.

In this mentorship, you will learn the `kuttl` tool, which is quite simple to use, and provide coverage for all the policies in the kyverno/policies repository (some of which have already been done). As part of your application, please see the Cover Letter section and be prepared to answer the challenge.

## Cover Letter

The Kyverno project receives many applicants for Linux Foundation mentorships. In order to select the most qualified and motivated candidate, in this mentorship we are asking you to read and respond to this challenge as a component of your cover letter. The response in your cover letter should not exceed more than one additional page in length.

Please also understand that this mentorship is **full time** in which the candidate is expected to deliver a 40-hour work week or show equivalent progress. If you cannot make this commitment, please do not apply.

### The Challenge

The `kuttl` tool is simple to learn and understand as it allows a declarative approach to testing of Kubernetes applications and does not require any programming language to write or understand tests. Tests are simply a collection of Kubernetes manifests and, sometimes, potentially some commands or very simple scripts when manifests alone are insufficient.

After spending a small amount of time with `kuttl`, review the many example tests [here](https://github.com/kyverno/kyverno/tree/main/test/conformance/kuttl) and propose a test for this Kyverno policy below. Use the `scaffolding` directory [here](https://github.com/kyverno/kyverno/tree/main/test/conformance/kuttl/_aaa_template_resources/scaffold) as template for the starter files necessary. If any additional test case files are needed, find them in its parent directory. Kyverno's fork of `kuttl` can be installed simply with `go install github.com/kyverno/kuttl/cmd/kubectl-kuttl@main`. The selected candidate will have the following as part of the cover letter's response.

1. The correct files in use.
2. A description of the test and why the test is needed.
3. Test cases, with an explanation, of the various angles needed to ensure that the policy works correctly in the majority of scenarios.
4. A link to a publicly-accessible git repo (GitHub suggested) containing all the files needed to test the cases.

```yaml
apiVersion: kyverno.io/v1
kind: ClusterPolicy
metadata:
[...]

PR #651, opened 2023-06-17
test: 7 kuttl tests from others

mintbomb27:
## Related Issue(s)
#546 

## Description
Contains 7 tests from others category.

## Checklist

### DIFF
--- a/other/allowed-image-repos/pod-good.yaml
+++ b/other/allowed-image-repos/pod-good.yaml
@@ -1,0 +1,35 @@
+apiVersion: v1
+kind: Pod
+metadata:
+  name: goodpod01
+spec:
+  containers:
+    - name: pod-01
+      image: myknownimage
+---
+apiVersion: v1
+kind: Pod
+metadata:
+  name: goodpod02
+spec:
+  containers:
+    - name: pod-01
+      image: myknownimage
+    - name: pod-02
+      image: kyverno
+---
+apiVersion: v1
+kind: Pod
+metadata:
+  name: goodpod03
+spec:
+  initContainers:
+    - name: pod-01-init
+      image: kyverno
+    - name: pod-02-init
+      image: myknownimage
+  containers:
+    - name: pod-01
+      image: myknownimage
+    - name: pod-02
+      image: kyverno
--- a/other/allowed-image-repos/pod-good.yaml
+++ b/other/allowed-image-repos/pod-good.yaml
@@ -14,9 +14,9 @@ metadata:
 spec:
   containers:
     - name: pod-01
-      image: myknownimage
+      image: ghcr.io/images/myknownimage:1.26
     - name: pod-02
-      image: kyverno
+      image: ghcr.io/kyverno/kyverno:latest
 ---
 apiVersion: v1
 kind: Pod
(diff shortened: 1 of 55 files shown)


other/allowed-image-repos/pod-good.yaml
```diff
@@ -0,0 +1,35 @@
+apiVersion: v1
+kind: Pod
+metadata:
+  name: goodpod01
+spec:
+  containers:
+    - name: pod-01
+      image: myknownimage
```
chipzoller, line 8: This policy only checks the image repository, not the tag/digest. I suggest adding tags to some (perhaps not all) images in all these test resources to better simulate an actual environment. The tag/digest does not matter.

other/allowed-label-changes/bad-update.yaml
chipzoller, line 1: It would be good in maybe one or two of these cases to try and change two labels where one includes the `breakglass` label. The expectation is this would also fail since something other than the allowed label has also changed.

other/audit-event-on-delete/04-assert.yaml
chipzoller, line 1: And you're finding, even without a `name`, this assertion works properly as well as consistently?

other/audit-event-on-exec/04-assert.yaml
chipzoller, line 1: Same with this?

chipzoller (changes requested):
Very good. Couple very minor things.

other/allowed-image-repos/pod-good.yaml
chipzoller: See also [here](https://kyverno.io/docs/writing-policies/variables/#variables-from-container-images) for an example of the fields that would get parsed out. Notice how `name` only includes the image name (i.e., repository).

mintbomb27: Ohh, will try that out.

other/audit-event-on-delete/04-assert.yaml
mintbomb27: Yes, the `name` can't be added because it is random, so it is confirmed with the other fields like the `involvedObject`. I confirmed this with the kuttl docs.

other/audit-event-on-exec/04-assert.yaml
mintbomb27: Yes, the other components like involvedObject, reason, and the source are asserted here, unlike other events happening in the namespace. This was checked manually as well.

other/audit-event-on-delete/04-assert.yaml
mintbomb27: @chipzoller Should I change the test in any way here?

other/allowed-label-changes/bad-update.yaml
mintbomb27: Added.

other/allowed-image-repos/pod-good.yaml
mintbomb27: I've added.

other/audit-event-on-delete/04-assert.yaml
chipzoller: No, this is good if you say it works.

other/audit-event-on-exec/04-assert.yaml
chipzoller: Ok

chipzoller (approved):
Nice work.

PR #653, opened 2023-06-19
test: 6 tests from others

mintbomb27:
## Related Issue(s)
#546 

## Description
Contains 6 tests from others

## Checklist

### DIFF
--- a/other/block-cluster-admin-from-ns/02-setup-user.yaml
+++ b/other/block-cluster-admin-from-ns/02-setup-user.yaml
@@ -1,0 +1,10 @@
+apiVersion: kuttl.dev/v1beta1
+kind: TestStep
+commands:
+- command: openssl genrsa -out testuser.key 2048
+- command: openssl req -new -key testuser.key -out testuser.csr -subj "/CN=testuser/O=testorg"
+- command: "docker cp kind-control-plane:/etc/kubernetes/pki/ca.crt ."
+- command: "docker cp kind-control-plane:/etc/kubernetes/pki/ca.key ."
+- command: openssl x509 -req -in testuser.csr -CA ca.crt -CAkey ca.key -CAcreateserial -out testuser.crt -days 10
+- command: kubectl config set-credentials testuser --client-certificate=testuser.crt --client-key=testuser.key
+- command: kubectl config set-context testuser-context --cluster=cluster --user=testuser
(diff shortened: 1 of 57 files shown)


other/block-cluster-admin-from-ns/07-delete.yaml
```diff
@@ -0,0 +1,6 @@
+apiVersion: kuttl.dev/v1beta1
+kind: TestStep
+commands:
+- command: kubectl delete -f good-pod.yaml
+- command: find . -name 'ca.*' -delete
+- command: find . -name 'testuser.*' -delete
```
chipzoller, line 6: No need to do this. Everything gets trashed at the end of the tests. The only things we care about cleaning up are things in the cluster which could skew/impact further tests that get run.

other/block-cluster-admin-from-ns/02-setup-user.yaml
```diff
@@ -0,0 +1,10 @@
+apiVersion: kuttl.dev/v1beta1
+kind: TestStep
+commands:
+- command: openssl genrsa -out testuser.key 2048
+- command: openssl req -new -key testuser.key -out testuser.csr -subj "/CN=testuser/O=testorg"
+- command: "docker cp kind-control-plane:/etc/kubernetes/pki/ca.crt ."
+- command: "docker cp kind-control-plane:/etc/kubernetes/pki/ca.key ."
```
chipzoller, line 7: These are fine, but I still prefer to eliminate as many tooling dependencies as we can, especially those which we're assuming to be on these runners. That's why in the script I gave you I read this info from the kubeconfig where it is already provided.

other/block-cluster-admin-from-ns/04-pods.yaml
```diff
@@ -0,0 +1,6 @@
+apiVersion: kuttl.dev/v1beta1
+kind: TestStep
+commands:
+- command: kubectl --context=testuser-context create -f good-pod.yaml
```
chipzoller, line 4: Also try a DELETE and UPDATE operation. Obviously, you'll want to create something first as the non `testuser` account, but it could be as simple as a ConfigMap.

chipzoller (commented):
Looks good, minor things.

other/block-cluster-admin-from-ns/02-setup-user.yaml
mintbomb27: Have changed to use the script now.

other/block-cluster-admin-from-ns/04-pods.yaml
mintbomb27: Done!

chipzoller (approved):

PR #658, opened 2023-06-21
test: block-pod-exec tests, stale-images, updates-deletes tests

mintbomb27:
## Related Issue(s)
#546 

## Description
Contains 6 kuttl tests for other/block-pod-exec-* and block-stale-images, block-updates-deletes

## Checklist

### DIFF
--- a/other/block-stale-images/01-assert.yaml
+++ b/other/block-stale-images/01-assert.yaml
@@ -1,0 +1,6 @@
+apiVersion: kyverno.io/v1
+kind: ClusterPolicy
+metadata:
+  name: block-stale-images
+status:
+  ready: true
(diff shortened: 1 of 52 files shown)


other/block-stale-images/01-assert.yaml
chipzoller, line 1: Ok, this one I'm torn over. The challenge we'll have is this test will start failing once the `busybox:1.35` image reaches 6 months of age, and then we'll have to find another tag to use :) My vote is to just skip this one so it doesn't require periodic refreshing.

chipzoller (approved):
All look good.

other/block-stale-images/01-assert.yaml
mintbomb27: What if I make it to use the `:latest` tag? Hopefully it will be maintained for a couple 10 years more?

chipzoller: Yes, let's do that.

chipzoller (approved):

PR #679, opened 2023-07-11
test: 7 tests from others

mintbomb27:
## Related Issue(s)
#546 

## Description
Contains 7 kuttl tests from others.

## Checklist

### DIFF
--- a/other/only-trustworthy-registries-set-root/pod-good.yaml
+++ b/other/only-trustworthy-registries-set-root/pod-good.yaml
@@ -1,0 +1,21 @@
+apiVersion: v1
+kind: Pod
+metadata:
+  name: goodpod01
+  namespace: trust-registry-ns
+spec:
+  containers:
+  - name: kyverno
+    image: ghcr.io/kyverno/kyverno:latest
+---
+apiVersion: v1
+kind: Pod
+metadata:
+  name: goodpod02
+  namespace: trust-registry-ns
+spec:
+  containers:
+  - name: kyverno01
+    image: ghcr.io/kyverno/kyverno:latest
+  - name: kyverno02
+    image: ghcr.io/kyverno/kyverno:latest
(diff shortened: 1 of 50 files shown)


other/only-trustworthy-registries-set-root/pod-good.yaml
```diff
@@ -0,0 +1,21 @@
+apiVersion: v1
+kind: Pod
+metadata:
+  name: goodpod01
+  namespace: trust-registry-ns
+spec:
+  containers:
+  - name: kyverno
+    image: ghcr.io/kyverno/kyverno:latest
```
chipzoller, line 9: This image does not run as root so you need an additional test here for an image which does run as root but comes from GHCR.

chipzoller:
```sh
crane config ghcr.io/kyverno/kyverno:latest | jq .config.User
"65532"
```

other/prevent-naked-pods/03-check.yaml
```diff
@@ -0,0 +1,5 @@
+apiVersion: kuttl.dev/v1beta1
+kind: TestStep
+commands:
+- command: sleep 5
+- script: if [ $(kubectl get pods -n prevent-naked-pods-ns | grep gooddeployment01 | wc -l) -gt 0 ]; then exit 0; else exit 1; fi
```
chipzoller, line 5: I thought you could assert on a partial name for a Pod, is that not so? You could just assert on the name `gooddeployment01-` (note the trailing dash character), if it does work, and it would be simpler. Otherwise, ignore this comment.

chipzoller (changes requested):
Good. Couple of things.

other/only-trustworthy-registries-set-root/pod-good.yaml
mintbomb27: Okay, adding `ghcr.io/nginx-proxy/nginx-proxy:latest`

other/prevent-naked-pods/03-check.yaml
mintbomb27: Tried this, didn't work. I wish it did though.

chipzoller:
Ready for re-review?

mintbomb27:
Yes @chipzoller

PR #680, opened 2023-07-12
test: 10 tests from others, k10-generate-gold-backup-policy

mintbomb27:
## Related Issue(s)
#546 

## Description
Contains 10 kuttl tests from others and 1 from kasten, `k10-generate-gold-backup-policy`

## Checklist


other/refresh-env-var-in-pod/04-check.yaml
chipzoller, line 1: This is an interesting way to go about the check. I suppose it works, but I would probably just make this simpler by doing a `kubectl exec` into the new Pod spawned when Kyverno annotates the Deployment to get value of the env var from the updated Secret. If the value is the same as the update made to the Secret, pass; otherwise, fail.

chipzoller: In order for what I suggested to work, those Deployments which run busybox need to perpetually sleep so they can be available for `exec`'ing later.

other/refresh-volumes-in-pods/04-check.yaml
chipzoller, line 1: Same type of comment here as in refresh-env-var-in-pod.

other/remove-hostpath-volumes/pods.yaml
```diff
@@ -0,0 +1,153 @@
...
+  - name: foo
+    hostPath:
+      path: /data
+  - name: bar
+    hostPath:
+      path: /data/junk
+  - name: empty
+    emptyDir:
+      medium: memory
+      sizeLimit: 20Mi
```
chipzoller, line 38: Swap the order here, please. We had a problem in the past where some logic wasn't working properly if these were separate. We want to ensure that doesn't happen again. Seems like a minor thing but it hurt us in the past.

other/remove-hostpath-volumes/remove-hostpath-volumes.yaml
```diff
@@ -48,7 +48,7 @@ spec:
                 op: remove
           - list: request.object.spec.containers[]
             foreach:
-            - list: element.volumeMounts
+            - list: " element.volumeMounts || `[]` "
```
chipzoller, line 51: Good catch on this, and good to see a test for it.

other/replace-image-registry/pods.yaml
```diff
@@ -0,0 +1,28 @@
...
+  initContainers:
+  - name: kyverno
+    image: kyverno:1.10.1
+  - name: busybox-init
+    image: docker.io/busybox:1.35
+  containers:
+  - name: busybox
+    image: registry.corp.com/kyverno:1.10.1
+  - name: busybox02
+    image: docker.io/busybox:1.35
```
chipzoller, line 28: Can you also add a longer path here, something like `ghcr.io/foo/bar/baz/busybox:1.35`?

chipzoller (changes requested):
Looks good. Just minor points.

kasten/k10-generate-gold-backup-policy/01-clusterrole.yaml
```diff
@@ -0,0 +1,18 @@
...
+kind: ClusterRole
+metadata:
+  name: kyverno:background-controller:k10-goldbackuppolicy
+  labels:
+    app.kubernetes.io/component: background-controller
+    app.kubernetes.io/instance: kyverno
+    app.kubernetes.io/part-of: kyverno
+rules:
+- apiGroups:
+  - '*'
```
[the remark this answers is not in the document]
mintbomb27: Really sorry! I just picked up the old test I'd written and checked if it works. Have changed this.

other/record-creation-details/record-creation-details.yaml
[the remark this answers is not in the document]
mintbomb27: Sure.

other/refresh-env-var-in-pod/04-check.yaml
mintbomb27: I tried doing this. This kind of involves some trouble as we have to deal with pod names. Once the update is made, the old deployment is terminated and the new deployment is fired, as that's how the `patchStrategicMerge` works. The time difference between this is kind of unstable. So we'll have to store the old pod name, then make the update to the secret, and then either wait for the merge to complete, or get the new pod name of the deployment and make the check. 

It's not impossible, but it might be more complicated than the current one. If you think the exec way is better, then I can look more into it!

other/remove-hostpath-volumes/pods.yaml
mintbomb27: Sure! Changing.

other/remove-hostpath-volumes/remove-hostpath-volumes.yaml
mintbomb27: Thanks!

other/replace-image-registry/pods.yaml
mintbomb27: I'm guessing in this case the regex should be `myregistry.corp.com/foo/bar/baz/busybox:1.35`. That is the output of the policy as well.

chipzoller: Yes

other/refresh-env-var-in-pod/04-check.yaml
chipzoller: Discussed this yesterday. Either approach will work.

other/refresh-volumes-in-pods/04-check.yaml
mintbomb27: For a not-watched configmap, when it is mounted in a pod, is it supposed to change the data with a change in the configmap automatically? So then how do I make this check? I've just checked whether the mounted data is what is expected.

chipzoller: If the ConfigMap is consumed as a volume, yes, but not as an env var. Does that answer this?

mintbomb27: Yes. So that's solved.

chipzoller (approved):

PR #681, opened 2023-07-12
test: kubeops, argo-cluster-generation-from-rancher-capi, consul

mintbomb27:
## Related Issue(s)
#546 

## Description
Contains test for:
* consul/enforce-min-tls-version
* kubeops/config-syncer-secret-generation-from-rancher-capi
* argo/argo-cluster-generation-from-rancher-capi

## Checklist


argo/argo-cluster-generation-from-rancher-capi/01-policy.yaml
```diff
@@ -0,0 +1,7 @@
+apiVersion: kuttl.dev/v1beta1
+kind: TestStep
+apply:
+- argo-cluster-generation-from-rancher-capi.yaml
```
chipzoller, line 4: As this is another "generate existing" policy, we should probably also have a step here which creates a Cluster resource first and then installs the policy. The Secret is expected to be generated both after the policy has been installed AND after a second Cluster resource has been created.

chipzoller: I also don't see a `00` step to assert the CRD here.

kubeops/config-syncer-secret-generation-from-rancher-capi/01-policy.yaml
```diff
@@ -0,0 +1,7 @@
+apiVersion: kuttl.dev/v1beta1
+kind: TestStep
+apply:
+- config-syncer-secret-generation-from-rancher-capi.yaml
```
chipzoller, line 4: Ditto here with respect to comments made on the Argo policy above.

chipzoller (changes requested):

argo/argo-cluster-generation-from-rancher-capi/01-policy.yaml
mintbomb27: I've added.

kubeops/config-syncer-secret-generation-from-rancher-capi/01-policy.yaml
mintbomb27: Done!

chipzoller (approved):

PR #693, opened 2023-07-22
test: 10 tests from others

mintbomb27:
## Related Issue(s)
#546 

## Description
Contains 10 kuttl tests from others

## Checklist


other/require-unique-external-dns/svc.yaml
```diff
@@ -0,0 +1,29 @@
+apiVersion: v1
+kind: Service
+metadata:
+  annotations:
+    external-dns.alpha.kubernetes.io/hostname: "1.1.1.1"
```
chipzoller, line 5: These aren't host names but that's ok, it still works.

other/restart-deployment-on-secret-change/06-check-restart.yaml
```diff
@@ -0,0 +1,4 @@
+apiVersion: kuttl.dev/v1beta1
+kind: TestStep
+commands:
+- script: "if [ \"$(kubectl get pods -n default | grep busybox | awk 'NR==1{print $1}')\" != \"$(cat pod-name.txt)\" ];then exit;else exit 1;fi"
```
chipzoller, line 4: You could have also compared the resourceVersion in the modified secret with the value of the `ops.corp.com/triggerrestart` but this is also ok :)

chipzoller (approved):
Nice

other/restrict-annotations/99-delete.yaml
```diff
@@ -0,0 +1,6 @@
+apiVersion: kuttl.dev/v1beta1
+kind: TestStep
+delete:
+- apiVersion: kyverno.io/v1
+  kind: ClusterPolicy
+  name: require-unique-uid-per-workload
```
chipzoller, line 6: Copy-and-paste error. The restrict-annotations policy isn't getting removed which is causing failures in the competing "allowed-annotations" test when it is run after this one.

chipzoller (changes requested):
Need to fix one error.

other/require-unique-external-dns/svc.yaml
mintbomb27: I'll change them 😅

other/restart-deployment-on-secret-change/06-check-restart.yaml
mintbomb27: Sounds like a better idea. I'll try this out.

mintbomb27: So I'm not sure if the annotation is getting removed after the restart, I'm not able to get the annotation `ops.corp.com/trigger`. It's just isn't there. But I can see the annotation `deployment.kubernetes.io/revision` getting incremented everytime a restart happens. Should I check with that?

other/restrict-annotations/99-delete.yaml
mintbomb27: Sorry about this! Fixed it!

other/restart-deployment-on-secret-change/06-check-restart.yaml
mintbomb27: I still had to save the `resourceVersion` of the old secret in a text file and then take it to compare with the annotation. But it was fun! Realized that the `resourceVersion` gets updated only after the kyverno admission check.

chipzoller (approved):

PR #696, opened 2023-07-24
Breaking `other` folder into smaller subfolders

prady0t:
## Related Issue(s)
Closes #688 

## Description

Made subfolders based on first names of the folders inside `other`. Those with the similar first name were moved to a new folder, rest were left as they were.  For example, `restrict-service-port-name` and `restrict-wildcard-resources` were moved inside the folder named `restrict`.

Test time was reduced from 13 min to around 4 minutes. 

Before:

Screenshot 2023-07-24 at 4 58 40 PM

After:
Screenshot 2023-07-24 at 9 35 40 PM

## Checklist

- [x] I have read the [policy contribution guidelines](https://github.com/kyverno/policies/blob/main/README.md#contribution).
- [] I have added test manifests and resources covering both positive and negative tests that prove this policy works as intended.
- [] I have added the artifacthub-pkg.yml file and have verified it is complete and correct.

PR #700, opened 2023-07-26
Splitting `other` folder into subfolders

### LINKED ISSUE #688: Break apart the `other` folder into smaller subfolders
The `other` folder has close to 100 policies which means kuttl tests of just that folder take quite some time. Find a strategy and break up those into smaller subfolders allowing tests to run in parallel per folder.

prady0t:
## Related Issue(s)
Closes #688 

## Checklist

***Note: Same PR as #696***


mintbomb27:
@prady0t Great work! But as you can see this doesn't bring any change to the time as you need to modify the workflow as well. The `Kuttl Tests` workflow still takes all the tests together from a folder and executes them. 

So would you mind changing the workflow as well to accommodate these changes? Basically, all you need to do I guess is in the `tests` array, instead of just `other`, you'll have to replace it with `other/a-b` and so on based on your folders.

prady0t:
@mintbomb27 Did changes in `test.yml` file. Please have a look.

.github/workflows/test.yml
```diff
@@ -42,7 +42,14 @@ jobs:
...
           - openshift
-          - other
+          # - other
+          - other/a
+          - other/b-d
+          - other/e-l
+          - other/m-p
+          - other/rec-req
+          - other/res
+          - other/s-v
```
chipzoller, line 52: Please rename this folder `s-z` both here in the workflow and the actual folder. Even though there aren't currently policies which begin with letters after `v`, this is where they would go.

chipzoller (changes requested):

mintbomb27 (approved):
LGTM! Thank you for the contribution!

chipzoller (approved):

chipzoller:
Thanks, @prady0t! Now we need to make similar adjustments to the workflow in kyverno/kyverno.

prady0t:
I can see folder `other` mentioned in `conformance.yaml` in kyverno/kyverno. Is it supposed to be `other/a` etc. just like the changes we made here? 
Do more such workflow files require such changes?

chipzoller:
That should be the only change.

prady0t:
Ok cool making a PR there them.👍🏻

chipzoller:
Thanks!

prady0t:
@chipzoller Have a look. https://github.com/kyverno/kyverno/pull/7927

PR #707, opened 2023-07-30
test: 11 tests from others

mintbomb27:
## Related Issue(s)
#546 

## Description
Contains 11 kuttl tests from others

## Checklist

### DIFF
--- a/other/res/restrict-node-label-changes/03-check.yaml
+++ b/other/res/restrict-node-label-changes/03-check.yaml
@@ -1,0 +1,7 @@
+apiVersion: kuttl.dev/v1beta1
+kind: TestStep
+commands:
+- script: |
+    node=$(kubectl get nodes --no-headers | awk 'NR==1{print $1}')
+    if kubectl label --overwrite nodes $node foo=foo; then echo "Failure: successfully modified foo"; exit 1; else echo "Success: failed to modify label foo"; fi
+    if kubectl label nodes $node foo-; then echo "Failure: successfully removed foo"; exit 1; else echo "Success: failed to remove label foo"; fi
--- a/other/res/restrict-node-label-changes/03-check.yaml
+++ b/other/res/restrict-node-label-changes/03-check.yaml
@@ -5,3 +5,6 @@ commands:
     node=$(kubectl get nodes --no-headers | awk 'NR==1{print $1}')
     if kubectl label --overwrite nodes $node foo=foo; then echo "Failure: successfully modified foo"; exit 1; else echo "Success: failed to modify label foo"; fi
     if kubectl label nodes $node foo-; then echo "Failure: successfully removed foo"; exit 1; else echo "Success: failed to remove label foo"; fi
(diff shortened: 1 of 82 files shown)


other/res/restrict-node-label-changes/03-check.yaml
```diff
@@ -0,0 +1,7 @@
+apiVersion: kuttl.dev/v1beta1
+kind: TestStep
+commands:
+- script: |
```
chipzoller, line 4: I would also like to see an addition, modification, and removal of another label aside from `foo`.

other/res/restrict-node-label-creation/03-check.yaml
```diff
@@ -0,0 +1,6 @@
+apiVersion: kuttl.dev/v1beta1
+kind: TestStep
+commands:
+- script: |
```
chipzoller, line 4: Same sort of test here. Make sure you can set some other label aside from `foo`.

other/res/restrict-pod-controller-serviceaccount-updates/03-manifests.yaml
chipzoller, line 1: How about some negative tests ensuring that another type of update to the `spec` object for a Deployment and a CronJob are allowed so long as it's not in the `serviceAccountName` field?

other/res/restrict-secret-role-verbs/cr-bad.yaml
chipzoller, line 1: Let's also try another "bad" resource in which only one of the forbidden verbs appears. That's what this policy checks.

other/res/restrict-secret-role-verbs/role-bad.yaml
chipzoller, line 1: Ditto here.

other/res/restrict-secrets-by-label/secret.yaml
chipzoller, line 1: Don't see a Secret here with a different label than `status` and no labels at all.

chipzoller (changes requested):

other/res/restrict-node-label-changes/03-check.yaml
mintbomb27: Okayy, added.

other/res/restrict-node-label-creation/03-check.yaml
mintbomb27: Added.

other/res/restrict-pod-controller-serviceaccount-updates/03-manifests.yaml
mintbomb27: Okay added.

other/res/restrict-secret-role-verbs/cr-bad.yaml
mintbomb27: Added.

other/res/restrict-secrets-by-label/secret.yaml
mintbomb27: Should have added that, sorry! Added now.

chipzoller (approved):
Nice!

PR #719, opened 2023-08-10
test: 4 tests from others

mintbomb27:
## Related Issue(s)
#546 

## Description
Contains the last 4 tests from others category. woohoo!

## Checklist


other/s-z/update-image-tag/00-crb.yaml
```diff
@@ -0,0 +1,20 @@
...
+  - apps
+  resources:
+  - deployments
+  verbs:
+  - update
+---
+apiVersion: v1
+kind: Namespace
+metadata:
+  name: update-image-tag-ns
```
chipzoller, line 20: Please break this out into a separate resource manifest.

other/s-z/update-image-tag/01-policy.yaml
```diff
@@ -0,0 +1,6 @@
+apiVersion: kuttl.dev/v1beta1
+kind: TestStep
+apply:
+- update-image-tag.yaml
```
chipzoller, line 4: Remember that this is a mutate existing policy and `spec.mutateExistingOnPolicyUpdate` here is set to `true`, which means we want to ensure that 1) the mutation occurs when the policy is created and 2) the mutation occurs when the policy is updated. Because you're creating the policy at the outside, you aren't testing for these scenarios.

other/s-z/update-image-tag/00-crb.yaml
mintbomb27: Sure! Done.

other/s-z/update-image-tag/01-policy.yaml
mintbomb27: How do you want me to update the policy? I can check for the first case, but second?

chipzoller: There's no update to the policy needed, there are test case updates needed.

mintbomb27: I've made the changes required.

PR #751, opened 2023-09-13
chore: remove kuttl deletion step

realshuting:


other/rec-req/require-base-image/99-delete.yaml
```diff
@@ -1,8 +0,0 @@
-apiVersion: kuttl.dev/v1beta1
-kind: TestStep
-delete:
-- apiVersion: kyverno.io/v1
-  kind: ClusterPolicy
-  name: require-base-image
-commands:
-- command: kubectl delete all --all --force --grace-period=0 -n require-base-image-ns
```
realshuting, line 8: Is the force deletion still required?

realshuting: It cased a timeout issue [here](https://github.com/kyverno/policies/actions/runs/6170604599/job/16747199526?pr=750) which led to a flaky test.

chipzoller: It's still needed in some very specific cases, but looking through this test case I don't think it's needed in this one.

chipzoller (approved):

chipzoller:
Let's confirm kuttl finishes successfully on this run.

chipzoller (changes requested):
We must retain this deletion step because the policy is created in a non-conventional way which sets `Enforce` mode using `sed`.

realshuting:
Seems like this change caused more failures. 

The cleanup task caused timeout, can we do them in separated steps? i.e., delete the pod first, wait 3s then delete the namespace?

chipzoller:
Yes, the cleanups can be done in different tasks, but we still have to delete the policy which is why this was causing failures.

PR #752, opened 2023-09-13
chore: split kuttl deletion tasks

realshuting:


realshuting:
Doesn't seem to help, closing.

PR #753, opened 2023-09-13
feat: add cri-dockerd flavor to disallow-cri-sock-mount policy

sp3nx0r:
## Related Issue(s)

N/A

## Description
During implementation in our environment, we noted that cri-dockerd was missing from the `disallow-cri-sock-mount` policy. So adding that simple use case in to the existing policy and updating associated test files. 

## Checklist

- [X] I have added the artifacthub-pkg.yml file and have verified it is complete and correct.

### DIFF
--- a/best-practices/disallow-cri-sock-mount/02-manifests.yaml
+++ b/best-practices/disallow-cri-sock-mount/02-manifests.yaml
@@ -9,6 +9,8 @@ apply:
   shouldFail: true
 - file: pod-crio-sock.yaml
   shouldFail: true
+- file: pod-cri-dockerd-sock.yaml
+  shouldFail: true
 - file: pod-emptydir-vol.yaml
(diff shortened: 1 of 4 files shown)


chipzoller:
Hi and thanks for this! Glad to accept it, only please also update the Artifact Hub metadata file as indicated [here](https://github.com/kyverno/policies#contribution).

chipzoller:
Also, please sign off on your git commits. This is a requirement of all CNCF projects.

chipzoller (changes requested):
No need to bump any other fields than `digest` with the new digest of the policy file which includes your changes.

sp3nx0r:
Hopefully that'll do it, sorry for the ignorance on Artifact Hub manipulation.

chipzoller (approved):
Thanks!

PR #756, opened 2023-09-20
fix: apply policies fixes

eddycharly:
## Related Issue(s)

This PR applies automatic policies fixes.


chipzoller:
Lots of kuttl failures I see

eddycharly:
I will need to dig this up !

eddycharly:
> changes to policies require recomputing the SHA256 hash and updating the corresponding Artifact Hub metadata file
Do we have an automatic validation for this ?

eddycharly:
@chipzoller i limited the changes to one folder for now

chipzoller (commented):
The problem with doing this is these changes are going to set the minimum compatibility to 1.11 which we can't do. The way the website as a front end to the policy library is set up to work is the rendering is done from `main` branch and the annotations are used to filter based on version.

I personally don't see a whole lot of value in this right now but I think the capability behind it is useful.

eddycharly:
@chipzoller why is this going to set the minimum compatibility to 1.11 ?

chipzoller:
You're adding the field `spec.admission` which only exists in 1.11.

eddycharly:
Wasn't this field introduced in 1.10 ?

Ok, this can easily be fixed.

chipzoller:
You're also using title case for `Audit` and `Enforce` which is a version pin to 1.9.

eddycharly:
Hmmm, let's think about it, we should be able to provide up to date policies.

chipzoller:
Two things that would make this more valuable: 1) be able to pass a Kyverno version and have policies updated to the language in that version and 2) run this against the website samples because for those we do always want them to be kept up-to-date with the current release.

PR #787, opened 2023-10-17
Update k10-3-2-1-backup.yaml

tdewin:
## Description

kubectl apply -n kyverno -f https://raw.githubusercontent.com/kyverno/policies/main/kasten/k10-3-2-1-backup/k10-3-2-1-backup.yaml
Warning: Validation failure actions enforce/audit are deprecated, use Enforce/Audit instead.

Changes audit to Audit

## Checklist


chipzoller:
Thanks for this, but we purposefully do not update all policies with the bleeding edge policy language. The reason being that it effectively pins the minimum version to the latest which disqualifies many of our users from consuming those policies, which increases the "support" load on us. These are warnings and can be ignored for now but should switch and use the preferred language later.

PR #794, opened 2023-10-26
Replace busybox:v1.35 with busybox:1.35

eskaaren:
## Related Issue(s)

N/A

## Description

Replace busybox:v1.35 with busybox:1.35
See https://hub.docker.com/_/busybox/tags?page=1

## Checklist

N/A

N/A


eskaaren:
I noticed the problem extends beyond the tag since the busybox image doesn't have a non root user set and will fail the kuttl tests where e.g. runAsNonRoot is set.

PR #800, opened 2023-11-03
fix(pdb-minavailable-check): ensure correct handling of 0 replicas workloads

erikgb:
## Related Issue(s)

Slack-thread: https://kubernetes.slack.com/archives/CLGR9BJU9/p1698916081905549

## Description

This PR adds another precondition to `pdb-minavailable-check` policy - to ensure that 0 replicas workloads are not denied by the policy. We added this policy to our clusters and experienced a lot of false-positives on workloads with `replicas: 0`, and no PDBs in namespace. I think this happens because the context variable defaults to 0.

## Checklist


chipzoller:
Thanks for this! We do need DCO sign-off on all commits though.

chipzoller (approved):

PR #806, opened 2023-11-11
fix the name field in artifacthub-pkg.yml for PSS CEL policies

MariamFahmy98:
## Related Issue(s)
None

## Description
CEL PSS policies aren't showing in Artifact Hub because of the duplicate names in the artifacthub-pkg.yml files. This PR adds `-cel` to the name.

## Checklist


chipzoller (approved):
Thanks!

PR #863, opened 2023-12-26
fix: use old busybox to fix digest changing

eddycharly:
Use old busybox to fix digest changing.


chipzoller (approved):
Not sure this will prevent the same thing from occurring since 1.35 was considered an "old tag" but approved.

eddycharly:
It looks like cli tests are not failing and they use 1.28 🤷 
At least it will make failures consistent between chainsaw and cli tests 😂

PR #873, opened 2024-01-02
fix: use conditional anchor instead of equality for restrict-scale

KhaledEmaraDev:
## Description

use conditional anchor instead of equality for restrict-scale

## Checklist


realshuting (approved):
Policy change looks good to me, needs @chipzoller's review on the artifacthub change.

chipzoller (approved):
Thank you

Issue #819, opened 2023-12-04
[Sample] policies to audit and restrict long-lived tokens for service accounts


JimBugwadia:
### Problem Statement

Prior to v1.24, Kubernetes generated long lived service accounts by default. These pose a security risk.

### Solution Description

Sample policies to:
* audit for secrets with `type: kubernetes.io/service-account-token`
* audit for service accounts with any `secrets.name` configured
* enforce pod `automountServiceAccountToken: false`

### Other Comments

See:
- https://kubernetes.io/docs/tasks/configure-pod-container/configure-service-account/
- https://kubernetes.io/docs/reference/access-authn-authz/service-accounts-admin/
- https://eng.d2iq.com/blog/service-account-tokens-in-kubernetes-v1.24/

### Slack discussion

https://kubernetes.slack.com/archives/CLGR9BJU9/p1701451535229319

### Troubleshooting

- [X] I have read and followed the documentation AND the [troubleshooting guide](https://kyverno.io/docs/troubleshooting/).
- [X] I have verified the policy does not exist in the [samples library](https://kyverno.io/policies/).

chipzoller:
Third request is already in the library [here](https://kyverno.io/policies/other/restrict-automount-sa-token/restrict-automount-sa-token/).

JimBugwadia:
@swastik959 - here is a policy using CEL that addresses the checks listed:

```yaml
apiVersion: kyverno.io/v1
kind: ClusterPolicy
metadata:
  name: deny-auto-mount-token
spec:
  validationFailureAction: Enforce
  background: false
  rules:
    - name: deny-auto-mount-token
      match:
        any:
        - resources:
            kinds:
              - Pod
      validate:
        cel:
          expressions:
            - message: "the default service account is not allowed"
              expression: >-
                object.spec.serviceAccountName != 'default'
            - message: "automountServiceAccountToken must be false" 
              expression: >-
                !has(object.spec.automountServiceAccountToken) || 
                object.spec.automountServiceAccountToken == false
            - message: "pod token expiration must be less than or equal to 1 hour"
              expression: >-
                !has(object.spec.volumes) || 
                  object.spec.volumes.all(v, !has(v.projected) || 
                    !has(v.projected.sources) ||
                      v.projected.sources.all(s, !has(s.serviceAccountToken) || 
                        !has(s.serviceAccountToken) || 
                        !has(s.serviceAccountToken.expirationSeconds) || 
[...]

PR #878, opened 2024-01-05
Added sample policy to check secret token type

siddhikhapare:
## Related Issue(s)
#819 

## Description
This policy aims to address  security problems by auditing to find and fix issues with certain types of long-lasting service accounts that could be risky.

## Checklist

### DIFF
--- a/other/audit-secret-token-type/audit-secret-token-type.yaml
+++ b/other/audit-secret-token-type/audit-secret-token-type.yaml
@@ -9,7 +9,14 @@ metadata:
     policies.kyverno.io/subject: Secret
     policies.kyverno.io/minversion: 1.6.0
     policies.kyverno.io/description: >-
-      Audit secrets of type kubernetes.io/service-account-token to ensure secure usage.
+      This policy audits for secrets of type 'kubernetes.io/service-account-token'.
+      In version v1.24, more enhancements have been introduced, affecting the behavior of 
+      service accounts and their secrets.In previous versions, when you create a service account, 
+      it automatically creates a secret resource associated with the service account.In version
+      v1.24 when you create a service account, secrets are no longer automatically created. 
+      To create a token,refer to the documentation for more details  
+      https://kubernetes.io/docs/tasks/configure-pod-container/configure-service-account/ and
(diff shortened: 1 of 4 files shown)


other/audit-secret-token-type/audit-secret-token-type.yaml
```diff
@@ -0,0 +1,42 @@
...
+  - name: validate-service-account-token-secrets
+    match:
+      any:
+      - resources:
+          kinds:
+          - Secret
+          namePattern: ".*" 
+          apiGroups: [""]
+          apiVersions: ["v1"]
+          resourceNames: [""]
```
chipzoller, line 33: Where do you get these from, ChatGPT? This isn't a valid policy at all and you would have known that if you tried to even create it, which also means you haven't personally tested it.

chipzoller (changes requested):
The policy isn't correct.

other/audit-secret-token-type/audit-secret-token-type.yaml
siddhikhapare: @chipzoller  I used chatgpt for reference but I thought using secrets  pattern will be helpful to get all resources such as pod and deployment to test type of that ```type: kubernetes.io/service-account-token```. I tested my files of resoure.yaml where I referred from various blogs  https://medium.com/@th3b3ginn3r/understanding-service-accounts-in-kubernetes-e9d2abe19df8 and https://spacelift.io/blog/kubernetes-secrets but I found error regarding this policy.

siddhikhapare: @chipzoller I tried with ```resource.object.type``` it's also not working. I will try with CEL.

chipzoller: Please do not use ChatGPT, or any other AI tool for that matter, to tell you what policies to write. Not only is it wrong, but it's just fantasy. Please use the docs, existing samples, and `kubectl explain` to understand the fields available when authoring policies. You do not need to resort to CEL to implement this simple policy.

siddhikhapare: @chipzoller Sure. Thank you for your feedback. I will look into it.

Issue #882, opened 2024-01-08
[Sample] rewrite image registry


JimBugwadia:
### Problem Statement

rewrite the image registry, to force all pulls throug a proxy cache e.g. with harbor

### Solution Description

```yaml
apiVersion: kyverno.io/v1
kind: ClusterPolicy
metadata:
  name: replace-image-registry
spec:
  background: false
  rules:
    - name: replace-image-registry
      match:
        any:
        - resources:
            kinds:
            - Pod
      mutate:
        foreach:
        - list: "request.object.spec.containers"
          context: 
          - name: imageData
            imageRegistry: 
              reference: "{{ element.image }}"
          preconditions:
            all:
            - key: "{{imageData.image}}"
              operator: Equals
              value: "docker.io/*"
          patchStrategicMerge:
            spec:
              containers:
              - name: "{{ element.name }}"
                image: "harbor2.zoller.com/dockerhub-proxy/{{imageData.repository}}:{{imageData.identifier}}"

```

### Slack discussion

https://kubernetes.slack.com/archives/CLGR9BJU9/p1704740196988469

### Troubleshooting

- [X] I have read and followed the documentation AND the [troubleshooting guide](https://kyverno.io/docs/troubleshooting/).
- [X] I have verified the policy does not exist in the [samples library](https://kyverno.io/policies/).

siddhikhapare:
I would like to work on this

jseiser:
We tried to use this, and it did not actually handle the bare image registry.  We ended up with this

```
    spec:
      background: false
      rules:
      - name: prepend-registry-containers
        match:
          any:
          - resources:
              kinds:
              - Pod
        preconditions:
          all:
          - key: "{{request.operation || 'BACKGROUND'}}"
            operator: AnyIn
            value:
            - CREATE
            - UPDATE
        mutate:
          foreach:
          - list: "request.object.spec.containers"
            patchStrategicMerge:
              spec:
                containers:
                - name: "{{ element.name }}"           
                  image: "harbor.dev.trex.network/{{ images.containers.\"{{element.name}}\".registry}}/{{ images.containers.\"{{element.name}}\".path}}:{{images.containers.\"{{element.name}}\".tag}}"
      - name: prepend-registry-initcontainers
        match:
          any:
          - resources:
              kinds:
              - Pod
        preconditions:
          all:
          - key: "{{request.operation || 'BACKGROUND'}}"
            operator: AnyIn
            value:
            - CREATE
            - UPDATE
[...]

PR #884, opened 2024-01-11
Updated replace image registry

siddhikhapare:
## Related Issue(s)
#882 

## Description
Rewrite the image registry, to force all pulls through a proxy cache with using harbor

## Checklist


siddhikhapare:
@JimBugwadia @chipzoller Could you please review my PR.

chipzoller (commented):
Why are you modifying the existing policy? I don't think we need to do that; I think the idea was to add a new one rather than modify the current one.

siddhikhapare:
> Why are you modifying the existing policy? I don't think we need to do that; I think the idea was to add a new one rather than modify the current one.

Sure. I'll make it happen. Thanks.

chipzoller:
Converting to draft. Please mark as ready for review when complete.

PR #898, opened 2024-02-02
Add disallow-proc-self-fd-workingdir.yaml (CVE-2024-21626)

lobuhi:
Mitigation for CVE-2024-21626

## Checklist

### DIFF
--- a/other/disallow-proc-self-fd-workingdir/artifacthub-pkg.yml
+++ b/other/disallow-proc-self-fd-workingdir/artifacthub-pkg.yml
@@ -3,8 +3,10 @@ version: 1.0.0
 displayName: Disallow workingDir with self file descriptors
 createdAt: "2024-02-02T00:00:00.000Z" 
 description: >-
-      In order to mitigate CVE-2024-21626, this policy prevent the creation of Pods
-      with workingDir set in /proc/self/fd/*
+      This policy prevents the configuration of a container workingDir with file descriptors
+      that start with /proc/self/fd. This is a partial mitigation for CVE-2024-21626 aka
+      Leaky Vessels Container Breakout. For additional details and mitigation please refer to
+      https://snyk.io/blog/leaky-vessels-docker-runc-container-breakout-vulnerabilities/.
 install: |-
     ```shell
     kubectl apply -f https://raw.githubusercontent.com/kyverno/policies/main/other/disallow-proc-self-fd-workingdir/disallow-proc-self-fd-workingdir.yaml
@@ -13,8 +15,10 @@ keywords:
   - runc
   - CVE-2024-21626
 readme: | 
-  In order to mitigate CVE-2024-21626, this policy prevent the creation of Pods
-  with workingDir set in /proc/self/fd/*
+  This policy prevents the configuration of a container workingDir with file descriptors
+  that start with /proc/self/fd. This is a partial mitigation for CVE-2024-21626 aka
+  Leaky Vessels Container Breakout. For additional details and mitigation please refer to
+  https://snyk.io/blog/leaky-vessels-docker-runc-container-breakout-vulnerabilities/.
 
   Refer to the documentation for more details on Kyverno annotations: https://artifacthub.io/docs/topics/annotations/kyverno/
 annotations: 
--- a/other/disallow-proc-self-fd-workingdir/artifacthub-pkg.yml
+++ b/other/disallow-proc-self-fd-workingdir/artifacthub-pkg.yml
@@ -1,0 +1,23 @@
+name: disallow-proc-self-fd-workingdir
+version: 1.0.0 
+displayName: Disallow workingDir with self file descriptors
+createdAt: "2024-02-02T00:00:00.000Z" 
+description: >-
+      In order to mitigate CVE-2024-21626, this policy prevent the creation of Pods
+      with workingDir set in /proc/self/fd/*
+install: |-
+    ```shell
+    kubectl apply -f https://raw.githubusercontent.com/kyverno/policies/main/other/disallow-proc-self-fd-workingdir/disallow-proc-self-fd-workingdir.yaml
+    ```   
+keywords:
+  - runc
+  - CVE-2024-21626
+readme: | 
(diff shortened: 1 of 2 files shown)


other/disallow-proc-self-fd-workingdir/disallow-proc-self-fd-workingdir.yaml
```diff
@@ -0,0 +1,35 @@
...
+      validate:
+        message: "Using /proc/self/fd in workingDir is not allowed."
+        pattern:
+          spec:
+            containers:
+              - =(workingDir): "!/proc/self/fd*"
+            initContainers:
+              - =(workingDir): "!/proc/self/fd*"
+            ephemeralContainers:
+              - =(workingDir): "!/proc/self/fd*"
```
chipzoller, line 35: These all need anchors.

other/disallow-proc-self-fd-workingdir/disallow-proc-self-fd-workingdir.yaml
```diff
@@ -0,0 +1,35 @@
...
+      that start with /proc/self/fd. This is a partial mitigation for CVE-2024-21626 aka
+      Leaky Vessels Container Breakout. For additional details and mitigation please refer to
+      https://snyk.io/blog/leaky-vessels-docker-runc-container-breakout-vulnerabilities/.
+spec:
+  validationFailureAction: enforce
+  background: true
+  rules:
+    - name: check-workingDir
+      match:
+        resources:
```
chipzoller, line 23: `any` block required

chipzoller (changes requested):
Please follow contribution guidelines in the PR template. An Artifact Hub metadata file, minimally, is required along with this PR.

other/disallow-proc-self-fd-workingdir/disallow-proc-self-fd-workingdir.yaml
lobuhi: Anchors added

other/disallow-proc-self-fd-workingdir/disallow-proc-self-fd-workingdir.yaml
lobuhi: Any block added

other/disallow-proc-self-fd-workingdir/disallow-proc-self-fd-workingdir.yaml
```diff
@@ -0,0 +1,39 @@
...
+          spec:
+            definitions:
+              customCheck: &customCheckAnchor
+                =(workingDir): "!/proc/self/fd*"
+            containers:
+              - <<: *customCheckAnchor
+            initContainers:
+              - <<: *customCheckAnchor
+            ephemeralContainers:
+              - <<: *customCheckAnchor
```
chipzoller, line 39: Sorry I wasn't clear on the anchors. I didn't mean traditional YAML anchors. The previous version of this was fine, but you'll need[ equality anchors](https://kyverno.io/docs/writing-policies/validate/#anchors) around `initContainers` and `ephemeralContainers` because a Pod may not have those and so the policy would fail for such Pods.

other/disallow-proc-self-fd-workingdir/artifacthub-pkg.yml
```diff
@@ -0,0 +1,27 @@
...
+  - CVE-2024-21626
+readme: | 
+  This policy prevents the configuration of a container workingDir with file descriptors
+  that start with /proc/self/fd. This is a partial mitigation for CVE-2024-21626 aka
+  Leaky Vessels Container Breakout. For additional details and mitigation please refer to
+  https://snyk.io/blog/leaky-vessels-docker-runc-container-breakout-vulnerabilities/.
+
+  Refer to the documentation for more details on Kyverno annotations: https://artifacthub.io/docs/topics/annotations/kyverno/
+annotations: 
+  kyverno/category: "Runc"
```
chipzoller, line 25:
```suggestion
  kyverno/category: "Security"
```

other/disallow-proc-self-fd-workingdir/disallow-proc-self-fd-workingdir.yaml
popeyeinit: Like this?
```yaml
            containers:
              - =(workingDir): "!/proc/self/fd*"
            =(initContainers):
              - =(workingDir): "!/proc/self/fd*"
            =(ephemeralContainers):
              - =(workingDir): "!/proc/self/fd*"
 ```

**Comment by chipzoller** [comment]
> I think the policy is not sufficient to prevent an attack, because the [WORKDIR](https://docs.docker.com/engine/reference/builder/#workdir) may be specified in the container image and not in the k8s manifest.

That's true and this should probably be specified in the description of the policy.

**Comment by popeyeinit** [comment]
> The WORKDIR instruction can resolve environment variables previously set using ENV. You can only use environment variables explicitly set in the Dockerfile.

And if I interpret the [docker docs](https://docs.docker.com/engine/reference/builder/#workdir) correctly, the `WORKDIR` maybe specified also via environment variables, so the policy should also match on that in k8s manifests?
```yaml
apiVersion: v1
kind: Pod
metadata:
  name: envar-demo
spec:
  containers:
  - name: envar-demo-container
    image: gcr.io/google-samples/node-hello:1.0
    env:
    - name: WORKDIR
      value: "/proc/self/fd/1"
```

popeyeinit:
> And if I interpret the [docker docs](https://docs.docker.com/engine/reference/builder/#workdir) correctly, the WORKDIR maybe specified also via environment variables, so the policy should also match on that in k8s manifests?

This is not correct. Although the environment variable `WORKDIR` may be set in an k8s manifest, it is not evaluated (picked up).

chipzoller:
This isn't going to work with the YAML anchors because Kyverno will interpret `spec.definitions.customCheck` as a literal path to the resource. Please re-write without using YAML anchors (but still using Kyverno validate anchors).

chipzoller:
@lobuhi, status?

chipzoller:
No response, closing.

PR #902, opened 2024-02-16
fix: require-unique-external-dns on updates

su-tiko:
## Description

Fixes an issue with `require-unique-external-dns` that was preventing updates to services using the external-dns annotation when in `Enforce` mode. This was caused because the API call to retrieve all the DNS names was not ignoring the request object current annotation, so it was always part of the list.

## Checklist


chipzoller (approved):

chipzoller:
Thank you

PR #907, opened 2024-02-22
chore: add 1.29 to the test grid and update kind images

eddycharly:
## Related Issue(s)

Add 1.29 to the test grid and update kind images.


realshuting (approved):
👍

PR #496, opened 2023-01-30
Add pdb checks

andrewhibbert:
## Description

Description of these policies:
* require-reasonable-pdbs
  * This checks PDBs and cross-references deployments/statefulsets/replicas if they are present
   * If percentages are used it will check that `minAvailable` is between 1 and 50% or `maxUnavailable` is between 50 and 100%
   * If numbers are used it will check that minAvailable is less than the number of corresponding replicas in a deployment/statefulset/replicaset. There is a precondition of replicas greater than 1
   * If numbers are used it will check that maxUnavailable does not equal 0
* require-replicas-allow-disruption
   * This checks deployments/statefulsets/replicas and cross references PDBs if they are available
   * If numbers are used it checks to see that replicas are not less than or equal to PDB `minAvailable`. There is a precondition of replicas greater than 1

## Checklist

- [x] I have read the [policy contribution guidelines](https://github.com/kyverno/policies/blob/main/README.md#contribution).
- [x] I have added test manifests and resources covering both positive and negative tests that prove this policy works as intended.

PR #920, opened 2024-03-03
New PDB policies

### LINKED ISSUE #801: [Sample] additional rule to validate PDB when workload exists in pdb-minavailable-check
### Problem Statement

We are having issues with users creating invalid PDB vs. workload in our multi-tenant clusters, and while the PDB policies in this policy catalog help, we have seen examples of PDBs added/updated:

- without any matching pods 
- matching existing workloads where the PDB is "invalid" on creation

The first case might be a bit difficult to handle, but the second one could be caught by a second rule in the existing [pdb-minavailable-check](https://github.com/kyverno/policies/blob/main/other/m-q/pdb-minavailable/pdb-minavailable.yaml) policy. The current rule assumes that the PDB already exists when the workload is created/updated. We want to make the policy also cover the opposite order: 

1. Workload with X replicas exists.
2. A PDB matching the workload is created. The PDB specifies `minAvailable: X` and should be denied by the policy.

### Solution Description

I would suggest adding a second rule to policy selecting on kind PDB, with the same overall validation as the existing rule. The most tricky part is probably how to find the desired number of replicas from the workload of controlling pods matching the PDB. Maybe we can use `ownerReferences`?

For the following good and bad resources, I will assume a workload already exists in the cluster. E.g.

````yaml
apiVersion: apps/v1
kind: StatefulSet
metadata:
  name: example
  namespace: pdb-minavailable-ns
spec:
  selector:
    matchLabels:
      app: busybox
  replicas: 2
  template:
    metadata:
      labels:
        app: busybox
    spec:
      terminationGracePeriodSeconds: 10
      containers:
      - name: busbyox
        image: busybox:1.35
````

### Example "Good" Resource

```yaml
apiVersion: policy/v1
kind: PodDisruptionBudget
metadata:
  name: ss-example-good
spec:
  minAvailable: 1
  selector:
    matchLabels:
      app: busybox
```

### Example "Bad" Resource

```yaml
apiVersion: policy/v1
kind: PodDisruptionBudget
metadata:
  name: ss-example-bad
spec:
  minAvailable: 2
  selector:
    matchLabels:
      app: busybox
```

### Other Comments

I would be happy to submit a PR to fix this, but I think I need some help on the policy logic - as I am new to Kyverno.

### Slack discussion

https://kubernetes.slack.com/archives/CLGR9BJU9/p1698916081905549

### Troubleshooting

- [X] I have read and followed the documentation AND the [troubleshooting guide](https://kyverno.io/docs/troubleshooting/).
- [X] I have verified the policy does not exist in the [samples library](https://kyverno.io/policies/).

chipzoller:
Signed-off-by: chipzoller <chipzoller@gmail.com>

## Related Issue(s)

Closes #801

## Description

Adds, expands, and improves new PodDisruptionBudget policies from PR #496.

* Ensure Deployment Replicas Higher Than PodDisruptionBudget
* PodDisruptionBudget maxUnavailable Non-Zero with Deployments
* Require Reasonable PodDisruptionBudgets
* Require Replicas Allow Disruption

## Checklist


chipzoller:
Would appreciate your review, @erikgb.

erikgb:
> @erikgb, any feedback?

It's on my list, @chipzoller. Sorry, but the PR is quite huge. 😉

other/require-replicas-allow-disruption/require-replicas-allow-disruption.yaml
```diff
@@ -0,0 +1,60 @@
...
+spec:
+  validationFailureAction: Audit
+  background: false
+  rules:
+  - name: replicas-check
+    match:
+      any:
+      - resources:
+          kinds:
+          - Deployment
```
erikgb, line 26: Should probably add StatefulSet as well. The structure in this area is the same, so it comes for "free".

other/require-replicas-allow-disruption/require-replicas-allow-disruption.yaml
```diff
@@ -0,0 +1,60 @@
...
+    validate:
+      message: >-
+        Replica count ({{ request.object.spec.replicas }}) cannot be less than or equal to the minAvailable of any
+        matching PodDisruptionBudget. There are {{ length(matchingpdbs) }} PodDisruptionBudgets which match this labelSelector,
+        not all of which may define a minAvailable value as a number.
+      foreach:
+      - list: matchingpdbs
+        preconditions:
+          all:
+          - key: '{{ regex_match(''^[0-9]+$'', ''{{ element.spec.minAvailable || ''''}}'') }}'
```
erikgb, line 52: Why is this precondition required? It also appears very complex. Can it be simplified, if required?

other/require-replicas-allow-disruption/.chainsaw-test/policy-ready.yaml
```diff
@@ -0,0 +1,6 @@
+apiVersion: kyverno.io/v1
+kind: ClusterPolicy
+metadata:
+  name: require-replicas-allow-disruption
+status:
+  ready: true
```
erikgb, line 6:
```suggestion
  ready: true

```
Nit

other/require-replicas-allow-disruption/.chainsaw-test/good-deploy.yaml
```diff
@@ -0,0 +1,20 @@
...
+      app: busybox
+  template:
+    metadata:
+      labels:
+        app: busybox
+    spec:
+      containers:
+      - image: busybox:1.36
+        name: busybox
+        command: ["sleep", "infinity"]
```
erikgb, line 20:
```suggestion
        command: ["sleep", "infinity"]

```

Nit

other/require-replicas-allow-disruption/.chainsaw-test/chainsaw-test.yaml
```diff
@@ -0,0 +1,35 @@
...
+    try:
+    - apply:
+        file: good-deploy.yaml
+  - name: 04 - Create bad Deployment
+    try:
+    - apply:
+        file: bad-deploy.yaml
+        expect:
+        - check:
+            ($error != null): true
```
erikgb, line 35:
```suggestion
            ($error != null): true

```

Nit

other/require-replicas-allow-disruption/.chainsaw-test/bad-deploy.yaml
```diff
@@ -0,0 +1,20 @@
...
+      app: busybox
+  template:
+    metadata:
+      labels:
+        app: busybox
+    spec:
+      containers:
+      - image: busybox:1.36
+        name: busybox
+        command: ["sleep", "infinity"]
```
erikgb, line 20:
```suggestion
        command: ["sleep", "infinity"]

```

Nit

other/require-reasonable-pdbs/require-reasonable-pdbs.yaml
```diff
@@ -0,0 +1,50 @@
...
+      # deny if minAvailable is greater than 50% or maxUnavailable is less than 50%
+      deny:
+        conditions:
+          any:
+          - key: '{{ regex_match(''^([1-9]|[1-4][0-9]|5[0])%$'', ''{{ request.object.spec.minAvailable || ''50%''}}'') }}'
+            operator: Equals
+            value: false
+          - key: '{{ regex_match(''^([5-9][0-9]|100)%$'', ''{{ request.object.spec.maxUnavailable || ''50%''}}'') }}'
+            operator: Equals
+            value: false
```
erikgb, line 50: This policy appears complex and very hard to read. Is it possible to simplify?

other/pdb-maxunavailable-with-deployments/pdb-maxunavailable-with-deployments.yaml
```diff
@@ -0,0 +1,56 @@
...
+    match:
+      any:
+      - resources:
+          kinds:
+          - PodDisruptionBudget
+          operations:
+          - CREATE
+          - UPDATE
+    context:
+    - name: deploymentreplicas
```
erikgb, line 32: If possible, I would suggest covering statefulsets as well with this policy.

other/pdb-maxunavailable-with-deployments/.chainsaw-test/policy-ready.yaml
```diff
@@ -0,0 +1,6 @@
+apiVersion: kyverno.io/v1
+kind: ClusterPolicy
+metadata:
+  name: pdb-maxunavailable-with-deployments
+status:
+  ready: true
```
erikgb, line 6:
```suggestion
  ready: true

```

Nit

other/pdb-maxunavailable-with-deployments/.chainsaw-test/chainsaw-test.yaml
```diff
@@ -0,0 +1,45 @@
...
+            name: alpha-busybox
+          spec:
+            replicas: 1
+  - name: 05 - Create bad PDBs
+    try:
+    - apply:
+        file: bad-pdb.yaml
+        expect:
+        - check:
+            ($error != null): true
```
erikgb, line 45:
```suggestion
            ($error != null): true

```

Nit

other/pdb-maxunavailable-with-deployments/.chainsaw-test/bad-pdb.yaml
```diff
@@ -0,0 +1,9 @@
+apiVersion: policy/v1
+kind: PodDisruptionBudget
+metadata:
+  name: bad-pdb01
+spec:
+  maxUnavailable: 0
+  selector:
+    matchLabels:
+      app: busybox
```
erikgb, line 9:
```suggestion
      app: busybox

```

Nit

other/deployment-replicas-higher-than-pdb/deployment-replicas-higher-than-pdb.yaml
```diff
@@ -0,0 +1,57 @@
...
+      PodDisruptionBudget resources are useful to ensuring minimum availability is maintained at all times.
+      Introducing a PDB where there are already matching Pod controllers may pose a problem if the author
+      is unaware of the existing replica count. This policy ensures that the minAvailable value is not
+      greater not equal to the replica count of any matching existing Deployment. If other Pod controllers
+      should also be included in this check, additional rules may be added to the policy which match those
+      controllers.
+spec:
+  validationFailureAction: Audit
+  background: true
+  rules:
```
erikgb, line 21: Could it make sense to merge the two policies addressing PDBs? For a user it is easy to remove one of the two rules - if required. And they are quite similar in nature.

other/deployment-replicas-higher-than-pdb/.chainsaw-test/policy-ready.yaml
```diff
@@ -0,0 +1,6 @@
+apiVersion: kyverno.io/v1
+kind: ClusterPolicy
+metadata:
+  name: deployment-replicas-higher-than-pdb
+status:
+  ready: true
```
erikgb, line 6:
```suggestion
  ready: true

```

Nit

other/require-replicas-allow-disruption/require-replicas-allow-disruption.yaml
chipzoller: Typically I like making the match as simple as possible to "illustrate" the use case and rely on the user to add others based on their specific needs. I don't mind adding just one other in this instance though.

other/require-reasonable-pdbs/require-reasonable-pdbs.yaml
chipzoller: Challenge here is the use of percentages. No Kyverno filters support operating on percentages. One potential way to simplify the deny conditions would be to strip off the percent sign (since the preconditions ensures it is present) and treat them as numbers.

other/deployment-replicas-higher-than-pdb/deployment-replicas-higher-than-pdb.yaml
chipzoller: We really try and keep policies separated as much as possible except in cases where there is such a high degree of relativity/coupling that one rule either doesn't make sense without the other or where users could potentially fall into a trap (due to them not knowing Kyverno well enough, or Kubernetes) if left to their own devices to find a complementary one. Given that these could be separate use cases I think it makes sense to separate them where the other advantage is a nice description for each.

other/require-replicas-allow-disruption/require-replicas-allow-disruption.yaml
tenstad: Would it make sense to put a space here, as there is a space in the begining? `''{{ element.`
```suggestion
          - key: '{{ regex_match(''^[0-9]+$'', ''{{ element.spec.minAvailable || '''' }}'') }}'
```

chipzoller:
And thank you for your review!

other/deployment-replicas-higher-than-pdb/deployment-replicas-higher-than-pdb.yaml
```diff
@@ -0,0 +1,57 @@
...
+      - resources:
+          kinds:
+          - PodDisruptionBudget
+          operations:
+          - CREATE
+          - UPDATE
+    context:
+    - name: deploymentreplicas
+      apiCall:
+        jmesPath: items[?label_match(`{{request.object.spec.selector.matchLabels}}`, spec.template.metadata.labels)] || `[]`
```
tenstad, line 34:
```suggestion
        jmesPath: items[?label_match(`{{ request.object.spec.selector.matchLabels }}`, spec.template.metadata.labels)] || `[]`
```
or is there a reson why there is no padding within the `{{ }}`?

chipzoller: Not particularly, just styling inconsistency.

other/deployment-replicas-higher-than-pdb/deployment-replicas-higher-than-pdb.yaml
erikgb: Makes sense! Thanks for the detailed response!

other/require-replicas-allow-disruption/require-replicas-allow-disruption.yaml
erikgb: Makes sense. Somehow it forces the user to understand the policy - which is good! 👍

other/deployment-replicas-higher-than-pdb/deployment-replicas-higher-than-pdb.yaml
```diff
@@ -0,0 +1,57 @@
...
+    - name: deploymentreplicas
+      apiCall:
+        jmesPath: items[?label_match(`{{ request.object.spec.selector.matchLabels }}`, spec.template.metadata.labels)] || `[]`
+        urlPath: /apis/apps/v1/namespaces/{{request.namespace}}/deployments 
+    preconditions:
+      all:
+      - key: '{{ regex_match(''^[0-9]+$'', ''{{ request.object.spec.minAvailable || ''''}}'') }}'
+        operator: Equals
+        value: true
+      - key: '{{ length(deploymentreplicas) }}'
```
erikgb, line 41: Is this precondition strictly required? I would expect the for loop below to deal with it, but maybe that's not how it works?

other/pdb-maxunavailable-with-deployments/pdb-maxunavailable-with-deployments.yaml
```diff
@@ -0,0 +1,56 @@
...
+      apiCall:
+        jmesPath: items[?label_match(`{{request.object.spec.selector.matchLabels}}`, spec.template.metadata.labels)] || `[]`
+        urlPath: /apis/apps/v1/namespaces/{{request.namespace}}/deployments
+    preconditions:
+      all:
+      - key: '{{ regex_match(''^[0-9]+$'', ''{{ request.object.spec.maxUnavailable || ''''}}'') }}'
+        operator: Equals
+        value: true
+      - key: '{{ length(deploymentreplicas) }}'
```
erikgb, line 41: Same comment as above. Is this strictly required - since there is a for loop below?

other/deployment-replicas-higher-than-pdb/deployment-replicas-higher-than-pdb.yaml
chipzoller: Maybe not technically, but from a reporting perspective yes because you want to differentiate between a `skip` and a `pass`. If there aren't any matching Deployments then this policy really isn't applicable, so exiting in the precondition is cleaner than going through the full policy evaluation.

other/require-reasonable-pdbs/require-reasonable-pdbs.yaml
erikgb: I see. Dealing with numbers is always easier. By naming a variable appropriately, it could be more readable/understandable. E.g. `maxUnavailablePercent`.

other/deployment-replicas-higher-than-pdb/deployment-replicas-higher-than-pdb.yaml
erikgb: Ah, of course! Thanks!

other/require-replicas-allow-disruption/require-replicas-allow-disruption.yaml
chipzoller: Done.

other/pdb-maxunavailable-with-deployments/pdb-maxunavailable-with-deployments.yaml
chipzoller: It could be done but it would have to be reworked with more context variables so the correct API call is made. I'll take this suggestion as a future enhancement request :)

other/require-reasonable-pdbs/require-reasonable-pdbs.yaml
chipzoller: I'll take this as another enhancement request. We'll see if these see any adoption!

erikgb (approved):
LGTM, but remember that I am a Kyverno noob, so this needs to be reviewed by someone with more experience. 😅

realshuting (approved):

PR #921, opened 2024-03-05
fix: initContainer ordering with server-side apply of image mutation

lexsca:
## Related Issue(s)

This is the closest issue I could find:

https://github.com/kyverno/kyverno/issues/3294#issuecomment-1080628983

## Description

This PR changes the patch method (of the policies I could find) to preserve the ordinality of `initContainers` when mutating their image location.

I'm submitting this PR in the hopes of saving others some pain around mutating the image location of init containers. Specifically, using the `patchStrategicMerge` method can scramble the order of init containers. Using `patchesJson6902` with indexing allows pods with multiple init containers to run in the expected order.

The extant tests seem to pass without any changes needed.

## Checklist

### DIFF
--- a/other/prepend-image-registry/artifacthub-pkg.yml
+++ b/other/prepend-image-registry/artifacthub-pkg.yml
@@ -19,4 +19,4 @@ annotations:
   kyverno/category: "Other"
   kyverno/kubernetesVersion: "1.21"
   kyverno/subject: "Pod"
-digest: 6325c3d888d0dcba78dcbe2c29f3fe6730addb8c4dda2e3c97b48ff8d3873943
+digest: 1d6897daf13c770973fb025516df1634232eb2e6cd4a38ece24a6af9b146294b
--- a/other/prepend-image-registry/prepend-image-registry.yaml
+++ b/other/prepend-image-registry/prepend-image-registry.yaml
@@ -59,8 +59,7 @@ spec:
     mutate:
       foreach:
       - list: "request.object.spec.initContainers"
-        patchStrategicMerge:
-          spec:
-            initContainers:
-            - name: "{{ element.name }}"           
-              image: registry.io/{{ images.initContainers."{{element.name}}".path}}:{{images.initContainers."{{element.name}}".tag}}
+        patchesJson6902: |-
+          - op: replace
+            path: /spec/initContainers/{{ elementIndex }}/image
+            value: registry.io/{{ images.initContainers."{{element.name}}".path}}:{{images.initContainers."{{element.name}}".tag}}
--- a/other/replace-image-registry/artifacthub-pkg.yml
+++ b/other/replace-image-registry/artifacthub-pkg.yml
@@ -19,4 +19,4 @@ annotations:
   kyverno/category: "Sample"
   kyverno/kubernetesVersion: "1.23"
   kyverno/subject: "Pod"
-digest: 9fde3b8caba3a05c5534b588fcd794975b8c1016004fa4ffdfb5bc4e93997e58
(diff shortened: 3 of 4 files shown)


chipzoller:
Looks like tests are failing.

chipzoller:
A couple of things here:
1. I'm not sure why you're finding it necessary in the first place to change the patch method for these two policies. In not only my tests but those in CI, the initContainer ordering is not changed with the current version of both policies. If you're seeing that is you likely have external influences at work, for example a service mesh which is performing the reordering.
2. I notice that the failure in the `replace-image-registry` policy is caused by the registry not being replaced when the patching method is switched from a strategic merge patch to a JSON patch. I can't presently explain it so I've logged https://github.com/kyverno/kyverno/issues/9878 for analysis.

lexsca:
> 1. I'm not sure why you're finding it necessary in the first place to change the patch method for these two policies. In not only my tests but those in CI, the initContainer ordering is not changed with the current version of both policies. If you're seeing that is you likely have external influences at work, for example a service mesh which is performing the reordering.

Changing the patch method was down to the confluence of a couple things:

a) I have a deployment with order-dependent initContainers required for pod startup. It worked fine locally on docker desktop k8s, but the initContainer order was scrambled in my EKS clusters. After some debugging, I discovered it was my Kyverno policy that re-writes the image location from docker hub to an ECR pull-through cache that caused the re-ordering. Removing the policy restored the initContainer order. I was also able to apply the same Kyverno policy locally to docker-desktop k8s and get the same scrambled initContainer order to happen.

b) I thought this behavior was maybe an edge-case in Kyverno, but then I came across [this snippet](https://kyverno.io/docs/writing-policies/tips/#mutate):

> Need to mutate an object at a specific ordered position within an array? Use the [patchesJson6902](https://kyverno.io/docs/writing-policies/mutate/#rfc-6902-jsonpatch) method.

Sure enough, switching the patch method from `patchStrategicMerge` to `patchesJson6902` and leaving everything else the same fixed the initContainer ordering and allowed the ECR pull-through cache image location re-writing to work as expected. 

I thought this would be an obvious change based on the Kyverno docs matching up with the observed behavior. Apologies for the confusion and lack of context opening this PR. Weirdly, I couldn't get the order to change with a Kyverno test case. It only manifested on a live k8s cluster. Server-side apply oddness maybe?

> 2. I notice that the failure in the `replace-image-registry` policy is caused by the registry not being replaced when the patching method is switched from a strategic merge patch to a JSON patch. I can't presently explain it so I've logged [[Bug] `foreach` patching behavior differs depending on method kyverno#9878](https://github.com/kyverno/kyverno/issues/9878) for analysis.

Interesting. FWIW, I'm using `patchesJson6902` in my image location re-writing policy right now and it's working as expected. I'll try to pull together a minimal reproducible example if that would be helpful.

chipzoller:
> I was also able to apply the same Kyverno policy locally to docker-desktop k8s and get the same scrambled initContainer order to happen.

Can you give me reproduction steps, including versions used and resources, on how you noticed this?

> I thought this behavior was maybe an edge-case in Kyverno, but then I came across [this snippet](https://kyverno.io/docs/writing-policies/tips/#mutate):

>> Need to mutate an object at a specific ordered position within an array? Use the [patchesJson6902](https://kyverno.io/docs/writing-policies/mutate/#rfc-6902-jsonpatch) method.

That note refers to a different use case. If you need to mutate, for example, the third (index 2) initContainer then you would need a JSON patch. In your case, you want to mutate _every_ initContainer that meets a certain criteria so this isn't quite the same thing as knowing the position you want to mutate.

lexsca:
i don't have the time to come up with a minimally viable reproducible test case on this, so closing for now.

Issue #824, opened 2023-12-04
[Sample] Add require-run-as-nonroot policy that use CEL expressions


MariamFahmy98:
### Problem Statement

Currently, all PSS policies are written in CEL expressions except the `require-run-as-nonroot` policy. We need to include it as well.  

Refer to [this](https://github.com/kyverno/policies/pull/776#issuecomment-1838426928) comment, for more information.

### Solution Description

N/A

### Troubleshooting

- [X] I have read and followed the documentation AND the [troubleshooting guide](https://kyverno.io/docs/troubleshooting/).
- [X] I have verified the policy does not exist in the [samples library](https://kyverno.io/policies/).

chipzoller:
The Slack thread in reference seems to be [here](https://kubernetes.slack.com/archives/C02TTBG6LF4/p1701685892590509), but I don't see any follow-up on your part. Would like to see how we can resolve and get this in with the objective being the complete set of PSS policies implemented as CEL subrules.

jayme-github:
I've re-added @MariamFahmy98 's policy from https://github.com/kyverno/policies/pull/776/commits/853bff9368a3304ad7f6fd8c936c4cbe3fa95132#diff-3416b7bb0c4cce31c5d5833e936ec552ea9d1426fadb6f2ec3b81bce4cb910a8 together with tests from `pod-security/restricted/require-run-as-nonroot/` and it passes all tests for me. Maybe this was just a glitch in kuttl?

PR #776, opened 2023-10-02
feat: add pod security standards (restricted) in CEL expressions - Part 3

MariamFahmy98:
## Related Issue(s)

Closes #773

## Description
This PR adds Kyverno policies written in CEL expressions for pod security standards (restricted).

## Checklist

- [x] I have read the [policy contribution guidelines](https://github.com/kyverno/policies/blob/main/README.md#contribution).
- [x] I have added test manifests and resources covering both positive and negative tests that prove this policy works as intended.
- [x] I have added the artifacthub-pkg.yml file and have verified it is complete and correct.

PR #923, opened 2024-03-06
Add require-run-as-nonroot policy that use CEL expressions

jayme-github:
## Related Issue(s)

Fixes: #824

## Description
Add require-run-as-nonroot policy using CEL expressions.
With the CEL from https://github.com/kyverno/policies/pull/776/commits/853bff9368a3304ad7f6fd8c936c4cbe3fa95132#diff-3416b7bb0c4cce31c5d5833e936ec552ea9d1426fadb6f2ec3b81bce4cb910a8 tests failed when Pod's `spec.securityContext` was  present but `spec.securityContext.runAsNonRoot` was not.

## Checklist


chipzoller (approved):
@MariamFahmy98, would you please review? I know this test also references resources outside of the parent directory, but we can fix that on all such CEL tests later.

MariamFahmy98 (approved):
LGTM.

Issue #956, opened 2024-03-25
[Chainsaw Tests] Add Chainsaw tests for the sample policy disallow-proc-mount


erisnar:
# Problem Statement

Add Chainsaw tests for Kyverno sample policy disallow-proc-mount

chipzoller:
Test for this isn't going to be possible. See comment [here](https://github.com/kyverno/policies/pull/957#issuecomment-2085397277).

PR #957, opened 2024-03-25
[Chainsaw Tests] Add Chainsaw tests for the sample policy disallow-proc-mount

erisnar:
## Related Issue(s)

#956 

## Description

Adds chainsaw tests to sample policy

## Checklist


chipzoller:
The tests for this are failing, the reason being, as noted in the description for this policy, Kubernetes will not allow any other procMount value to be set other than `Default` unless a corresponding feature gate is set. Since the clusters used to perform these tests do not set this feature gate, and not do we want to, writing this test isn't going to be possible (which is why no Chainsaw test was included by default).

Issue #891, opened 2024-01-19
Convert `Kubernetes Best Practices` policies to CEL


JimBugwadia:
### Problem Statement

Kubernetes Best Practices policies are written using Kyverno patterns and JMESPath, which means they cannot be executed as ValidatingAdmissionPolicy resources in the API server.

### Solution Description

Convert Kubernetes Best Practices policies, and other validating policies, to CEL wherever possible.

### Troubleshooting

- [X] I have read and followed the documentation AND the [troubleshooting guide](https://kyverno.io/docs/troubleshooting/).
- [X] I have verified the policy does not exist in the [samples library](https://kyverno.io/policies/).

swastik959:
Seems interesting will to applying to it .

coder12git:
@MariamFahmy98 @JimBugwadia  @anusha94  just for confirmation here, i wanna to ask that with `Kubernetes Best Practices policies`, you mean policies that are present inside this [best-practices](https://github.com/kyverno/policies/tree/main/best-practices) folder and this issue states to convert those policies into CEL? Am i right here?

chipzoller:
@MariamFahmy98, you've got a lot of PRs to review here 😄

Chandan-DK:
There's more on the way :)

PR #961, opened 2024-04-07
feat: add other policies in CEL expressions - Part 3

Chandan-DK:
## Related Issue(s)
Partially addresses #891 

## Description
This PR includes the conversion of policies in the other folder to Kyverno CEL policies. Conversion of the policies is done in multiple PRs

**Policies converted in this PR:**

- [x] memory-requests-equal-limits
- [x] metadata-match-regex
- [x] pdb-maxunavailable
- [x] prevent-bare-pods
- [x] prevent-cr8escape

## Checklist


other-cel/metadata-match-regex/.chainsaw-test/chainsaw-test.yaml
```diff
@@ -0,0 +1,38 @@
...
+    - patch:
+        resource:
+          apiVersion: kyverno.io/v1
+          kind: ClusterPolicy
+          metadata:
+            name: metadata-match-regex
+          spec:
+            validationFailureAction: Enforce
+    - assert:
+        file: chainsaw-step-01-assert-1.yaml
```
Chandan-DK, line 22: Reminder comment to rename `chainsaw-step-01-assert-1.yaml` files to `policy-ready.yaml`

Chandan-DK: Done.

other-cel/memory-requests-equal-limits/memory-requests-equal-limits.yaml
```diff
@@ -0,0 +1,38 @@
...
+      cel:
+        variables:
+          - name: containersWithResources
+            expression: object.spec.containers.filter(container, has(container.resources))
+        expressions:
+        - expression: >-
+            variables.containersWithResources.all(container, 
+            !has(container.resources.requests) ||
+            !has(container.resources.requests.memory) ||
+            container.resources.requests.memory == container.resources.?limits.?memory.orValue('-1'))
```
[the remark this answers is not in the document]
Chandan-DK: In this logic we check if `requests.memory` is equal to `limits.memory` or not.

We have 2 cases here:

1. `requests.memory` is not set and  `limits.memory` is either set or not set.

The logic should evaluate to true because `requests.memory` defaults to the value of `limits.memory` .
Here, is the description of `requests`.
```
$ k explain pod.spec.containers.resources.requests
KIND:       Pod
VERSION:    v1

FIELD: requests <map[string]Quantity>

DESCRIPTION:
    Requests describes the minimum amount of compute resources required. If
    Requests is omitted for a container, it defaults to Limits if that is
    explicitly specified, otherwise to an implementation-defined value. 
```

We use `!has(container.resources.requests) || !has(container.resources.requests.memory)` to account for this default behavior and return true.

2. `requests.memory` is set

i)  `limits.memory` is not set.

The logic should evaluate to false. And to do this, we have used the optional syntax `?` with `orValue()`.
So, in `container.resources.?limits.?memory.orValue('-1')`, if any of the fields `limits`, `memory` is not specified, the default value `-1` specified in `orValue() ` is compared with `requests.memory`, which will cause the logic to return a false, because `requests.memory` will surely not be the value -1 as it is invalid.
```
$ k apply -f -<<EOF
> apiVersion: v1
kind: Pod
metadata:
  name: badpod01
spec:
  containers:
  - name: busybox
    image: busybox:1.35
    resources:
      requests:
        memory: "-1"
> EOF
The Pod "badpod01" is invalid: spec.containers[0].resources.requests[memory]: Invalid value: "-1": must be greater than or equal to 0
```

ii) `limits.memory` is set.
In this case, the actual value of  `limits.memory` is directly compared with `requests.memory` and the `orValue('-1')` is not considered.

Not using the optional syntax can make the expression complex. 
Please let me know if it is clear.

MariamFahmy98 (approved):
LGTM

PR #967, opened 2024-04-11
Added test for copy-ns-labels policy

### LINKED ISSUE #950: [Chainsaw Tests] Add Chainsaw tests for the sample policies
### Problem Statement

Hi there👋

This is an umbrella issue for adding Chainsaw tests for Kyverno sample policies.

You can find references below to write tests:

- [Chainsaw doc](https://kyverno.github.io/chainsaw/latest/)
- [Example tests](https://github.com/kyverno/kyverno/tree/main/test/conformance/chainsaw)

Start by browsing the following sample policies, and you can pick up a task by:
1. create a GitHub issue on kyverno/policies repo
2. comment `/assign` to get assigned
3. write your tests
4. verify tests locally
  - require a Kubernetes cluster
  - require installing Chainsaw
5. create the PR

- [ ] https://github.com/kyverno/policies/tree/main/aws/require-aws-node-irsa
- [ ] https://github.com/kyverno/policies/tree/main/best-practices/check-deprecated-apis
- [x] refactor https://github.com/kyverno/policies/tree/main/cert-manager/limit-dnsnames
- [x] refactor https://github.com/kyverno/policies/tree/main/cert-manager/limit-duration
- [x] refactor https://github.com/kyverno/policies/tree/main/cert-manager/restrict-issuer
- [ ] https://github.com/kyverno/policies/tree/main/cleanup/cleanup-bare-pods
- [ ] https://github.com/kyverno/policies/tree/main/cleanup/cleanup-empty-replicasets
- [ ] https://github.com/kyverno/policies/tree/main/other/advertise-node-extended-resources
- [ ] https://github.com/kyverno/policies/tree/main/other/copy-namespace-labels
- [ ] https://github.com/kyverno/policies/tree/main/other/expiration-for-policyexceptions
- [ ] https://github.com/kyverno/policies/tree/main/other/namespace-protection
- [ ] https://github.com/kyverno/policies/tree/main/other/restrict-edit-for-endpoints
- [ ] https://github.com/kyverno/policies/tree/main/other/restrict-scale
- [ ] https://github.com/kyverno/policies/tree/main/other/restrict-service-account
- [ ] https://github.com/kyverno/policies/tree/main/pod-security/baseline/disallow-proc-mount
- [ ] https://github.com/kyverno/policies/tree/main/pod-security/subrule/podsecurity-subrule-baseline

### Solution Description

n/a

### Example "Good" Resource

_No response_

### Example "Bad" Resource

_No response_

### Other Comments

_No response_

### Slack discussion

_No response_

### Troubleshooting

- [X] I have read and followed the documentation AND the [troubleshooting guide](https://kyverno.io/docs/troubleshooting/).
- [X] I have verified the policy does not exist in the [samples library](https://kyverno.io/policies/).

siddhikhapare:
## Related Issue(s)

Partially addresses #950 

## Description
added chainsaw test for [copy-namespace-labels](https://github.com/kyverno/policies/tree/main/other/copy-namespace-labels) policy to test working of policy for outside and within ns with owner and multiple additional labels.

## Checklist


siddhikhapare:
@chipzoller Could you help me in resolving this issue of test failures?

chipzoller:
Sure. Forget the failures for right now. I see you are missing some fundamental points of what this policy does.
**Hint:** Look at what the policy matches on versus what you're testing. Do they all make sense?

siddhikhapare:
> Sure. Forget the failures for right now. I see you are missing some fundamental points of what this policy does. **Hint:** Look at what the policy matches on versus what you're testing. Do they all make sense?

Thank you for your helpful response. As per my understanding ,in order to validate both conditions of policy simultaneously. I have to define both rules in single resource. Could you please share your thoughts on this?

chipzoller:
Not sure what you're saying. Look at my hint, then look at the policy match block, then look at all the resources you're trying to create as part of the test. Something is off there. What is it?

siddhikhapare:
> Not sure what you're saying. Look at my hint, then look at the policy match block, then look at all the resources you're trying to create as part of the test. Something is off there. What is it?

pod resource kind is not included in the match block which I have provided and rules will only apply to deployment resources which will create and manage the associated pods.

siddhikhapare:
@chipzoller I will update it.

chipzoller:
Yes, correct. Fix that first then let's look.

siddhikhapare:
Should use a ClusterRole to grant permissions here for this policy?

chipzoller:
There should be no need to grant permissions here.

siddhikhapare:
@chipzoller I've noticed that policy is taking more time to initialize. I tried increasing the sleep duration in `chainsaw-test.yaml` file but still providing me same error regarding policy status is ``null`` which I mentioned [here](https://live.staticflickr.com/65535/53661716587_0cb0e6caf3_b.jpg). Could you please help me in understanding what's causing this issue?

siddhikhapare:
I assert that `owner` label is not a static value, so we do not have to specify in ``values.yaml``.
`owner` label value is fetched dynamically from the namespace's metadata labels using the `apiCall` and `jmesPath` in the policy's context.

chipzoller (changes requested):
Please look at the failed test log fro your error. You can and should also run this locally on your machine and you will see the same thing. You should always be running tests locally before pushing changes in a PR.

siddhikhapare:
> Please look at the failed test log fro your error. You can and should also run this locally on your machine and you will see the same thing. You should always be running tests locally before pushing changes in a PR.

I have tested my files using kyverno test previously. I have attached passed test output [here](https://live.staticflickr.com/65535/53669490335_3ce8f0df99_b.jpg) using chainsaw.

siddhikhapare:
@chipzoller  I tested files locally which have got output as passed tests as expected but here test are failed.

chipzoller:
Look at the logged output in CI tests here and see why they fail.

siddhikhapare:
for bad-deployment-with-ouside-ns resource have the `kubernetes.io/metadata.name` label. According to the policy, this label will not be copied to the deploy. However, other labels from the  ns should  be copied to the deployments. Therefore, this resource should fail because I didn't copy the owner label from the ns to the deploy before. It [failed](https://live.staticflickr.com/65535/53670676697_b41d03ee20_b.jpg) locally once because the API call is taking time to run, but it [passed](https://live.staticflickr.com/65535/53671910469_8ba9a4eb8e_b.jpg) the test once so I removed that resource.

siddhikhapare:
I think the tests were failing because of the `creationTimestamp` field, so I removed it . But when I ran the tests on locally, I found that they all passed, whether I included that field or not. You can see the results here [output](https://live.staticflickr.com/65535/53672653922_b565ed9c60_b.jpg
)

chipzoller:
You still don't seem to understand what this policy is about and how to test it. This is a mutation policy with two rules. There is no "bad" resource and no failure should occur. The test is about asserting on what the resulting mutation should look like, for both rules, in different scenarios with applicable resources.

siddhikhapare:
> You still don't seem to understand what this policy is about and how to test it. This is a mutation policy with two rules. There is no "bad" resource and no failure should occur. The test is about asserting on what the resulting mutation should look like, for both rules, in different scenarios with applicable resources.

I found [internal error](https://live.staticflickr.com/65535/53672885032_216c1374a8_b.jpg
) today that failed when calling the webhook "mutate-policy.kyverno.svc". However, according to this note mentioned in Kyverno's documentation https://kyverno.io/docs/writing-policies/mutate/ , Kubernetes disallows changes to certain fields in resources including `name, namespace, uid, kind, and apiVersion`, therefore you cannot use Kyverno policy to effect any of these fields either during admission or once a resource has been persisted. I thought I needed to change the `creationTimestamp` field, which indicates when an object was created. and this also mentioned in k8s doc here https://kubernetes.io/docs/concepts/workloads/pods/#pod-update-and-replacement so - Most of the metadata about a Pod is immutable that cannot change the `namespace, name, uid, or creationTimestamp fields`; the generation field is unique. It only accepts updates that increment the field's current value. This means that these fields can't be modified once they're set. 

I did not understand that error correctly so thanks for providing clarification.

chipzoller:
None of what you're saying matters in this case. A mutation during admission is not subject to these limitations since there is no "existing" resource. New resources being created are mutated before being persisted. All that's needed is to check the resource after it is created.

siddhikhapare:
All [test](https://live.staticflickr.com/65535/53679719537_7e14df85d5_b.jpg) passed as you have explained @chipzoller. Thanks for your guidance.

siddhikhapare:
I think this test might be showing there's a issue with resources timeout. Could you please provide your insights regarding this issue it would be helpful for me.

other/copy-namespace-labels/.chainsaw-test/resource.yaml
```diff
@@ -0,0 +1,85 @@
...
+        image: busybox:1.35
+---
+#Deployment with `kubernetes.io/metadata.name` label
+apiVersion: apps/v1
+kind: Deployment
+metadata:
+  name: my-deployment-2
+  namespace: within-ns
+  labels:
+    kubernetes.io/metadata.name: "my-deployment-2"
```
chipzoller, line 30: This shouldn't be here. Please remove. The label `kubernetes.io/metadata.name` is only found at the Namespace level.

chipzoller (changes requested):
Please also use an error step to fail if the `kubernetes.io/metadata.name` is found on at least one Deployment.

siddhikhapare:
@chipzoller Sorry for delay in resolving issue. I was out of station. I have checked test here that `kubernetes.io/metadata.name` label for only one [resource](https://live.staticflickr.com/65535/53708386060_3c10efa1a9_b.jpg). I have checked for all resource definition also to matched that label and they should provide error - ```- error:
        resource:
          apiVersion: apps/v1
          kind: Deployment
          metadata:
            labels:
            
  namespace: within-ns
              
[kubernetes.io/metadata.name](http://kubernetes.io/metadata.name): within-ns``` for this I found this [output](https://live.staticflickr.com/65535/53708398535_c27409a1b4_b.jpg)

chipzoller:
So are you saying the previous comment is resolved? Ready for review again?

Also, in the future, there's no need to take a screenshot of terminal output. Please just paste the text into a code block here so it's easy to follow the path of communication.

siddhikhapare:
> kubernetes.io/metadata.name

yes I believe that all test passed for that change what you have suggested to add.

siddhikhapare:
I have added whole resource definition in test file if that kubernetes.io/metadata.name label found with value of the label is the namespace name then that resource should provide error.

JimBugwadia:
@siddhikhapare - has this requested change been addressed? 

https://github.com/kyverno/policies/pull/967#pullrequestreview-2031487506

siddhikhapare:
@JimBugwadia @chipzoller I have added [here](https://github.com/kyverno/policies/pull/967/commits/8e0c5492c380cb6729de295847a18a7537e7f113#diff-777abbfd7c3c06fd5fe0ec73c3a4b0284a58ea755e5b09046ecc9b1b44b33910R21) as whole deployment resource to check that label present or not because for all resources contain immutable label kubernetes.io/metadata.name which set by Kubernetes control plane on all namespaces so that resource error should found. I have tested that file locally, it passed test.

other/copy-namespace-labels/.chainsaw-test/patchresource.yaml
```diff
@@ -0,0 +1,91 @@
...
+  template:
+    metadata:
+      labels:
+        app: good-app
+    spec:
+      containers:
+      - name: good-app-deploy
+        image: busybox:1.35
+---
+#Deployment with `kubernetes.io/metadata.name` label
```
chipzoller, line 24: Comment is no longer applicable.

other/copy-namespace-labels/.chainsaw-test/patchresource.yaml
```diff
@@ -0,0 +1,91 @@
...
+apiVersion: apps/v1
+kind: Deployment
+metadata:
+  name: my-deployment-2
+  namespace: within-ns
+  labels:
+    owner: "any-corp"
+    env: dev
```
chipzoller, line 32: You're missing the `app` label from the source Deployment. You must assert that it is present as well.

other/copy-namespace-labels/.chainsaw-test/resource.yaml
```diff
@@ -0,0 +1,85 @@
...
+        image: busybox:1.35
+---
+#Deployment with only `owner` label
+apiVersion: apps/v1
+kind: Deployment
+metadata:
+  name: my-deployment-3
+  namespace: within-ns
+  labels:
+    owner: "any-corp"
```
chipzoller, line 52: You should change the value of this label and ensure it is overwritten with the label's value from the Namespace. The patched resource for this is fine, but change the value of `owner` here to something different.

chipzoller (changes requested):

other/copy-namespace-labels/.chainsaw-test/chainsaw-test.yaml
```diff
@@ -0,0 +1,43 @@
...
+    - error:
+        resource:
+          apiVersion: apps/v1
+          kind: Deployment
+          metadata:
+            namespace: within-ns
+            name: my-deployment-2
+            labels:
+              kubernetes.io/metadata.name: within-ns
+              app: my-app02
```
chipzoller, line 30: You don't need to error on `app`, only on `kubernetes.io/metadata.name`.

chipzoller (changes requested):

siddhikhapare:
@chipzoller Should I update the digest hash string in the `artifacthub-pkg.yml` file? I believe it should be updated when the policy definition changes. I'm not sure what the exact lint error is. Could you help me understand it better?

siddhikhapare:
I found this error  diff --git a/kasten/kasten-generate-policy-by-preset-label/artifacthub-pkg.yml b/kasten/kasten-generate-policy-by-preset-label/artifacthub-pkg.yml

chipzoller:
Sent #1036 to fix.

chipzoller (approved):
Nice work, thanks.

PR #971, opened 2024-04-18
Update deprecated validationFailureAction value (audit/enforce)

mveitas:
## Description
Policies that are documented on the Kyverno site are using a deprecated values, `audit` and `enforce` for `validationFailureAction` attribute in the policies. The correct values should be `Audit` and `Enforce`

## Checklist

### DIFF
--- a/README.md
+++ b/README.md
@@ -59,7 +59,7 @@ metadata:
     policies.kyverno.io/description: >-
       Adding capabilities beyond those listed in the policy must be disallowed.
 spec:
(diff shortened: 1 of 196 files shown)


chipzoller:
Thanks for the contribution. We're leaving these put for right now by design. We use annotations to express minimum compatibility with the intention of scraping via the website. By updating these, it will change that value (which hasn't been reflected here). We'll probably bump these up next version.

PR #1018, opened 2024-05-19
Disallow images without Digest

nsagark:
## Related Issue(s)

This is Kyverno equivalent policy for below checkov policy. Using image digest in the deployment manifest is a security best practice. 
https://github.com/bridgecrewio/checkov/blob/main/checkov/kubernetes/checks/resource/k8s/ImageDigest.py

## Checklist


chipzoller (changes requested):
We already have a policy for this and so would prefer you to expand on it rather than add a duplicate.

https://kyverno.io/policies/other/require-image-checksum/require-image-checksum/

PR #1019, opened 2024-05-19
Require CPU Limits

nsagark:
## Related Issue(s)

This is Kyverno equivalent policy for below. Setting CPU limits on pods ensures fair distribution of resources, preventing any single pod from monopolizing CPU and impacting the performance of other pods. This practice enhances stability, predictability, and cost control, while also mitigating the noisy neighbor problem and facilitating efficient scaling in Kubernetes clusters. This policy ensures that cpu limits are set on every container.

https://github.com/bridgecrewio/checkov/blob/main/checkov/kubernetes/checks/resource/k8s/CPULimits.py

## Checklist


chipzoller (changes requested):
Once again, we have a policy for this here:  https://kyverno.io/policies/best-practices/require-pod-requests-limits/require-pod-requests-limits/

In practice, Kubernetes will set requests equal to limits if no requests are specified, so this policy is functionally equivalent to the above.

This PR also lumps in the policy from your previous PR and should not.

chipzoller:
I see I was mistaken. We don't have one which specifically requires CPU limits. Although the claim you make in your post isn't exactly correct, we can add such a policy to the library if you address issues.

chipzoller:
DCO is required here.

chipzoller:
Please sign off on your PR.

chipzoller:
Sign off is not complete.

chipzoller:
Converting to draft until all basic requirements are met.

PR #1022, opened 2024-05-20
Cleanup rs with timebound

anuddeeph1:
## Description
ReplicaSets serve as an intermediate controller for various Pod controllers like Deployments. When a new version of a Deployment is initiated, it generates a new ReplicaSet with the specified number of replicas and scales down the current one to zero. Consequently, numerous empty ReplicaSets may accumulate in the cluster, leading to clutter and potential false positives in policy reports if enabled. This cleanup policy is designed to remove empty ReplicaSets across the cluster within a specified timeframe, for instance, ReplicaSets created one day ago, ensuring the ability to rollback to previous ReplicaSets in case of deployment issues

## Checklist


chipzoller:
Can we not just update the policy [here ](https://kyverno.io/policies/cleanup/cleanup-empty-replicasets/cleanup-empty-replicasets/)instead of creating another one with only one additional expression?

PR #1025, opened 2024-05-21
Update restrict-wildcard-verbs to handle null/empty rules

danielreed:
## Description

Use the jmespath `to_array` function to better handle nulls on the request.object.rules[] path. In the case of a null a null will be added to an array so the rule can run without error.

## Checklist


chipzoller (changes requested):
Thanks for your contribution. Please see the contributing guide for all steps required when updating a policy.

chipzoller:
DCO is required here.

chipzoller (changes requested):
Apologies for not catching this on a previous review. Any time a policy is updated, the digest must be re-computed for the sake of Artifact Hub.

chipzoller (approved):

PR #1042, opened 2024-05-31
feat: add best practices check-deprecated-apis policy in CEL expressions 

Chandan-DK:
## Related Issue(s)
Partially addresses #891 

## Description
PR to convert `check-deprecated-apis` policy in the `best-practices` folder to CEL

## Checklist


best-practices-cel/check-deprecated-apis/check-deprecated-apis.yaml
```diff
@@ -0,0 +1,95 @@
...
+          - policy/*/PodDisruptionBudget
+          - policy/*/PodSecurityPolicy
+          - node.k8s.io/*/RuntimeClass
+    celPreconditions:
+      - name: "allowed-api-versions"
+        expression: "object.apiVersion in ['batch/v1beta1', 'discovery.k8s.io/v1beta1', 'events.k8s.io/v1beta1', 'policy/v1beta1', 'node.k8s.io/v1beta1']"
+    validate:
+      cel:
+        expressions:
+          - expression: "false"
```
MariamFahmy98, line 40: The original policy uses the `deny` conditions and I think the deny behavior can't be achieved through CEL expressions. What do you think about using the expression defined in CEL preconditions instead?

Chandan-DK: I'm not sure how we can use it in CEL preconditions in this case, because `cel.expressions` would become empty.

I just tried this out:

Here's a regular policy with `deny: {}`:
```yaml
apiVersion: kyverno.io/v1
kind: ClusterPolicy
metadata:
  name: restrict-operations-on-pod
spec:
  validationFailureAction: Enforce
  background: true
  rules:
  - name: rule-1
    match:
      any:
      - resources:
          kinds:
          - Pod
    validate:
      message: Create, update and delete on pods is not allowed
      deny: {} 
```

This policy blocks creating, updating and deleting pods

Here's the CEL policy for the above and it makes use of `expression: "false"`:

```yaml
apiVersion: kyverno.io/v1
kind: ClusterPolicy
metadata:
  name: restrict-operations-on-pod
spec:
  validationFailureAction: Enforce
  background: true
  rules:
  - name: rule-1
    match:
      any:
      - resources:
          kinds:
          - Pod
    validate:
      cel:
        expressions:
          - expression: "false"
            message: Create, update and delete on pods is not allowed
```

This policy blocks creating and updating pods. But it does not block deleting pods unlike the original policy. I wonder if this would be considered a bug?
Otherwise, I think we should be able to mimic the behaviour of `deny` using `expression: "false"`

Chandan-DK: Also while testing with a minimal version of this policy, I realized we encounter this bug (https://github.com/kyverno/kyverno/issues/10313).

```
cdk:~/exp$ kind create cluster --image kindest/node:v1.26.14
Creating cluster "kind" ...
 ✓ Ensuring node image (kindest/node:v1.26.14) 🖼
 ✓ Preparing nodes 📦  
 ✓ Writing configuration 📜 
 ✓ Starting control-plane 🕹️ 
 ✓ Installing CNI 🔌 
 ✓ Installing StorageClass 💾 
Set kubectl context to "kind-kind"
You can now use your cluster with:

kubectl cluster-info --context kind-kind

Not sure what to do next? 😅  Check out https://kind.sigs.k8s.io/docs/user/quick-start/
```

```
cdk:~/exp$ helm install -n kyverno kyverno kyverno/kyverno --create-namespace --version 3.2.1
NAME: kyverno
LAST DEPLOYED: Thu Jun 20 11:41:43 2024
NAMESPACE: kyverno
STATUS: deployed
REVISION: 1
NOTES:
Chart version: 3.2.1
Kyverno version: v1.12.1

Thank you for installing kyverno! Your release is named kyverno.

The following components have been installed in your cluster:
- CRDs
- Admission controller
- Reports controller
- Cleanup controller
- Background controller


⚠️  WARNING: Setting the admission controller replica count below 2 means Kyverno is not running in high availability mode.

💡 Note: There is a trade-off when deciding which approach to take regarding Namespace exclusions. Please see the documentation at https://kyverno.io/docs/installation/#security-vs-operability to understand the risks.
```

```
cdk:~/exp$ cat check-deprecated-apis 
apiVersion: kyverno.io/v1
kind: ClusterPolicy
metadata:
  name: check-deprecated-apis
spec:
  validationFailureAction: Audit
  background: true
  rules:
  - name: validate-v1-27-removals
    match:
      any:
      - resources:
          kinds:
          - storage.k8s.io/*/CSIStorageCapacity
    celPreconditions:
      - name: "allowed-api-versions"
        expression: "object.apiVersion in ['storage.k8s.io/v1beta1']"
    validate:
      cel:
        expressions:
          - expression: "false"
            messageExpression: >-
              object.apiVersion + '/' + object.kind + ' is deprecated and will be removed in v1.27.
              See: https://kubernetes.io/docs/reference/using-api/deprecation-guide/'
```
```
cdk:~/exp$ k apply -f check-deprecated-apis 
Error from server: error when creating "check-deprecated-apis": admission webhook "validate-policy.kyverno.svc" denied the request: no unique match for kind CSIStorageCapacity
```

We would have to wait for the issue to be resolved before considering merging this PR

MariamFahmy98: Have you filed an issue for it in the Kyverno repo?

Chandan-DK: I have logged an issue for not being able to replicate `deny: {}` behaviour using `expression: "false"`: [Link](https://github.com/kyverno/kyverno/issues/10576)

MariamFahmy98 (approved):
LGTM

Chandan-DK:
@MariamFahmy98 We probably have to wait for this issue (https://github.com/kyverno/kyverno/issues/10313) to be resolved before merging this PR. This policy won't function properly as of now (Comment link for the details: https://github.com/kyverno/policies/pull/1042#discussion_r1647560658)

PR #1057, opened 2024-06-29
[Chainsaw Tests] Add Chainsaw test for Bare Pods

Jay179-sudo:
## Related Issue(s)
Partial fix for #950 

## Description
Added chainsaw test for bare pods. Also added a test pod and a cluster-role manifest file.

## Checklist


Jay179-sudo:
Maybe we can also reduce the scheduled time on the ClusterCleanupPolicy file from five minutes down to one minute to reduce CI time?
Please do let me know if any other issues persist.

realshuting:
> Maybe we can also reduce the scheduled time on the ClusterCleanupPolicy file from five minutes down to one minute to reduce CI time? Please do let me know if any other issues persist.

Makes sense, can you replicate the same policy and reduce the scheduled time to a minute?

cleanup/cleanup-bare-pods/cleanup-bare-pods.yaml
```diff
@@ -28,4 +28,4 @@ spec:
     - key: "{{ target.metadata.ownerReferences[] || `[]` }}"
       operator: Equals
       value: []
-  schedule: "*/5 * * * *"
+  schedule: "*/1 * * * *"
```
realshuting, line 31: Not sure if it's ok to update the original policy, I was thinking to replicate the policy in the test folder and update the schedule.

@chipzoller - is this policy used anywhere else?

realshuting (approved):

chipzoller (changes requested):
Please move the ClusterRole inside the `.chainsaw-test` directory.

cleanup/cleanup-bare-pods/.chainsaw-test/cleanup-bare-pods-test.yaml
chipzoller, line 1: The policy should not be duplicated like this even if the need is to just modify a field. Please use an appropriate step in the Chainsaw test to modify the `schedule` field of the existing policy.

chipzoller (changes requested):

cleanup/cleanup-bare-pods/clusterrole.yaml
```diff
@@ -0,0 +1,20 @@
...
+  labels:
+    app.kubernetes.io/component: cleanup-controller
+    app.kubernetes.io/instance: kyverno
+    app.kubernetes.io/part-of: kyverno
+  name: kyverno:cleanup-controller:temp
+rules:
+- apiGroups:
+  - ""
+  resources:
+  - "pods"
```
chipzoller, line 14:
```suggestion
  - pods
```

cleanup/cleanup-bare-pods/clusterrole.yaml
```diff
@@ -0,0 +1,20 @@
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRole
+metadata:
+  namespace: kyverno
+  labels:
+    app.kubernetes.io/component: cleanup-controller
+    app.kubernetes.io/instance: kyverno
+    app.kubernetes.io/part-of: kyverno
+  name: kyverno:cleanup-controller:temp
```
chipzoller, line 9:
```suggestion
  name: kyverno:cleanup-controller:barepods
```

Jay179-sudo:
@chipzoller any updates on this?

cleanup/cleanup-bare-pods/.chainsaw-test/clusterrole.yaml
mohamedawnallah, line 1: Rename this file to `cluster-role.yaml` to be consistent with other files in the codebase?

mohamedawnallah (approved):
Great work @Jay179-sudo! 💪 Left some nit comments...

cleanup/cleanup-bare-pods/.chainsaw-test/chainsaw-step-02-assert-1.yaml
```diff
@@ -0,0 +1,4 @@
+apiVersion: kyverno.io/v2beta1
+kind: ClusterCleanupPolicy
+metadata:
+  name: clean-bare-pods
```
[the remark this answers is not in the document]
Jay179-sudo: The status.ready field is not defined for a ClusterCleanupPolicy object. It is defined within status.conditions.

When I ran the tests as normal, the status field that does get filled up is the lastExecutionTime field. However, this is an field that only gets filled up after a round of execution which will not allow for direct condition checks in our chainsaw test file.

Wouldn't it be better to skip this field altogether?

Status Fields during execution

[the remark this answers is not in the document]
mohamedawnallah: Yes you are right 👍

mohamedawnallah (approved):

cleanup/cleanup-bare-pods/.chainsaw-test/chainsaw-test.yaml
```diff
@@ -0,0 +1,38 @@
+apiVersion: chainsaw.kyverno.io/v1alpha1
+kind: Test
+metadata:
+  name: cleanup-bare-pods
+spec:
+  steps:
+    - name: apply clusterrole
+      try:
+        - apply:
+            file: clusterrole.yaml
```
mohamedawnallah, line 10: @Jay179-sudo It seems you forgot to rename this as well to `cluster-role.yaml`.

PS:
Also the DCO signing

chipzoller (approved):

PR #1062, opened 2024-07-05
Included 'jamespath' in the message to enable this user to view the container name in the policy report.

anuddeeph1:
## Description

Any capabilities added beyond those specified in the policy must be prohibited. Additionally, ensure that 'containername' is included in the policy report.

## Checklist


chipzoller (commented):
I don't really think this is best approach to getting the desired result. I think the better approach is to re-write this as a `foreach` so that way you can use the `message` field in a condition to reflect the element name.

chipzoller:
Converting to draft. Otherwise, feel free to close and open anew.

chipzoller:
This PR has been in draft for a month. Is there plans to complete? If not, please close.

PR #1066, opened 2024-07-09
Add initContainers and ephemeralContainers to Require Images Use Checksums

nsagark:
## Checklist


chipzoller (changes requested):
You don't need to add `name` as a conditional since this is a mandatory field. You also don't need to deviate from the pattern shown since digests can be specified using algorithms other than sha256. Please just add in ephemeral and initContainers following the same pattern that has already been established.

chipzoller (changes requested):
And if you're adding in checks for additional containers, you should cover for them in the test cases which doesn't look like has been done.

nsagark:
@chipzoller I have updated the policy. Also, I have updated the chainsaw tests to include init containers. I could not include ephermeral container as my chainsaw test was failing with below error.. Please review and let me know. 

` Pod "goodpod04" is invalid: spec.ephemeralContainers: Forbidden: cannot be set on create
    | 22:20:58 | require-image-checksum | step-02  | APPLY     | ERROR | v1/Pod @ chainsaw-mint-sponge/goodpod04
        === ERROR
        Pod "goodpod04" is invalid: spec.ephemeralContainers: Forbidden: cannot be set on create`

chipzoller:
You cannot add ephemeral containers on a create request. These can only be added on an update. Either use `kubectl debug` in a step or apply an updated manifest which contains them.

nsagark:
Hi @chipzoller Do we have any policy examples where we have chainsaw test case written for ephemeral containers. I tried something like below but getting an error. 

`# yaml-language-server: $schema=https://raw.githubusercontent.com/kyverno/chainsaw/main/.schemas/json/test-chainsaw-v1alpha1.json
apiVersion: chainsaw.kyverno.io/v1alpha1
kind: Test
metadata:
  creationTimestamp: null
  name: require-image-checksum
spec:
  steps:
  - name: step-01
    try:
    - apply:
        file: ../require-image-checksum.yaml
    - patch:
        resource:
          apiVersion: kyverno.io/v1
          kind: ClusterPolicy
          metadata:
            name: require-image-checksum
          spec:
            validationFailureAction: Enforce
    - assert:
        file: chainsaw-step-01-assert-1.yaml
  - name: step-02
    try:
    - apply:
        file: pod-good.yaml
    - apply:
        expect:
        - check:
            ($error != null): true
        file: pod-bad.yaml
    - apply:
        file: podcontroller-good.yaml
    - apply:
        expect:
        - check:
            ($error != null): true
        file: podcontroller-bad.yaml
  - name: step-03
    try:
    - apply:
        file: pod-good.yaml
    - patch:
        resource:
          apiVersion: v1
          kind: Pod
          metadata:
            name: goodpod01
          spec:
            containers:
            - name: busybox
              image: busybox@sha256:67a8ef886e2ca4055f00e7cd13aedb9b24148c1451a6832d16fcc997a157eedc
            - name: busybox02
              image: busybox@sha256:67a8ef886e2ca4055f00e7cd13aedb9b24148c1451a6832d16fcc997a157eedc
            ephemeralContainers:
            - name: debug-container
              image: busybox@sha256:67a8ef886e2ca4055f00e7cd13aedb9b24148c1451a6832d16fcc997a157eedc
              command: ['sh', '-c', 'sleep 1000']
---     
`

`   | 12:52:43 | require-image-checksum | step-03  | PATCH     | ERROR | v1/Pod @ chainsaw-vital-doberman/goodpod01
        === ERROR
        Pod "goodpod01" is invalid: spec: Forbidden: pod updates may not change fields other than `spec.containers[*].image`,`spec.initContainers[*].image`,`spec.activeDeadlineSeconds`,`spec.tolerations` (only additions to existing tolerations),`spec.terminationGracePeriodSeconds` (allow it to be set to 1 if it was previously negative)
          core.PodSpec{
                Volumes:        {{Name: "kube-api-access-vhvdt", VolumeSource: {Projected: &{Sources: {{ServiceAccountToken: &{ExpirationSeconds: 3607, Path: "token"}}, {ConfigMap: &{LocalObjectReference: {Name: "kube-root-ca.crt"}, Items: {{Key: "ca.crt", Path: "ca.crt"}}}}, {DownwardAPI: &{Items: {{Path: "namespace", FieldRef: &{APIVersion: "v1", FieldPath: "metadata.namespace"}}}}}}, DefaultMode: &420}}}},
                InitContainers: nil,
                Containers: []core.Container{`

chipzoller:
Please use code blocks to show code/YAML. I can't read that.

nsagark:
I have edited my response. Please take a look.

chipzoller:
You can't patch with `containers[]` being as part of the partial resource. Try with only ephemeralContainers.

nsagark:
Tried with just ephemeral containers. But still seeing an error. 


```
apiVersion: chainsaw.kyverno.io/v1alpha1
kind: Test
metadata:
  creationTimestamp: null
  name: require-image-checksum
spec:
  steps:
  - name: step-01
    try:
    - apply:
        file: ../require-image-checksum.yaml
    - patch:
        resource:
          apiVersion: kyverno.io/v1
          kind: ClusterPolicy
          metadata:
            name: require-image-checksum
          spec:
            validationFailureAction: Enforce
    - assert:
        file: chainsaw-step-01-assert-1.yaml
  - name: step-02
    try:
    - apply:
        file: pod-good.yaml
    - apply:
        expect:
        - check:
            ($error != null): true
        file: pod-bad.yaml
    - apply:
        file: podcontroller-good.yaml
    - apply:
        expect:
        - check:
            ($error != null): true
        file: podcontroller-bad.yaml
  - name: step-03
    try:
    - apply:
        file: pod-good.yaml
    - patch:
        resource:
          apiVersion: v1
          kind: Pod
          metadata:
            name: goodpod01
          spec:
            ephemeralContainers:
            - name: debug-container
              image: busybox@sha256:67a8ef886e2ca4055f00e7cd13aedb9b24148c1451a6832d16fcc997a157eedc
              command: ['sh', '-c', 'sleep 1000']
---     
```

Here is the error:

```
| 13:11:57 | require-image-checksum | step-03  | PATCH     | ERROR | v1/Pod @ chainsaw-humorous-quagga/goodpod01
        === ERROR
        Pod "goodpod01" is invalid: spec: Forbidden: pod updates may not change fields other than `spec.containers[*].image`,`spec.initContainers[*].image`,`spec.activeDeadlineSeconds`,`spec.tolerations` (only additions to existing tolerations),`spec.terminationGracePeriodSeconds` (allow it to be set to 1 if it was previously negative)
```

chipzoller:
Ok, the issue is that ephemeral containers is a subresource, so you can't patch the Pod directly. My suggestion would be to do a `kubectl debug` as a Chainsaw step.

nsagark:
Hi @chipzoller I have updated the chainsaw-test.yaml to include tests for ephemeral containers as well. Please take a look.

other/require-image-checksum/.kyverno-test/resource.yaml
chipzoller, line 1: Please remove ephemeralContainers from these resources. As mentioned, they are not valid at creation. Even though the CLI doesn't care, all test resources need to be valid.

chipzoller (changes requested):
Almost there.

nsagark:
Hi @chipzoller I have removed the ephemeral containers from the resource.yaml. Please take a look.

chipzoller (approved):

PR #1069, opened 2024-07-09
"Signed-off-by: Vikash Kaushik <vikash@nirmata.com>"

vikashkaushik01:
## Checklist


chipzoller (changes requested):
Appreciate the contribution here, but this doesn't seem like the right approach to prevent deploying the Kubernetes dashboard. Matching on all (`"*"`) kinds is very specialized and should be used only very carefully, and further matching just on a string in the name to prevent a behavior isn't usually reliable.

I'd suggest a different implementation here which users can view as more reliable.

chipzoller:
Any follow up here?

chipzoller:
Closed. No response.

PR #1071, opened 2024-07-10
added policy to replace nodeSelector

anuddeeph1:
## Description
The nodeSelector field, which uses labels to select the node for Pod scheduling based on specific needs, is updated by this policy to replace the label node-role.kubernetes.io/worker-abc with node.kubernetes.io/worker-abc in the Pod spec.

## Checklist


chipzoller (changes requested):
I don't understand the value of such a policy and your description does not explain it either. Can you explain to us why this is valuable for users? What does this provide/demonstrate that is useful?

anuddeeph1:
> I don't understand the value of such a policy and your description does not explain it either. Can you explain to us why this is valuable for users? What does this provide/demonstrate that is useful?

@chipzoller, the usecase is bit different, i will update according and create a new request, when the policy is ready

PR #1075, opened 2024-07-11
Remove hardcoded test namespace from restrict-volume-types tests

jayme-github:
## Description

pod-secuity policies used to hardcode the test namespace as restrict-voltypes-ns to be able to properly delete test resources. This change removes the need to hardcode the namespace name by using the build in namespace binding.

## Checklist
  - not applicable
  - not applicable


jayme-github:
`chainsaw/other/block-large-images` from `E2E Tests / v1.25 - ^other$/^[b-d]` fails, but I don't see how that should be related to this change.

chipzoller (approved):

PR #1080, opened 2024-07-19
Add policy Ensure HPA for Deployments

anuddeeph1:
## Description
This policy ensures that Deployments are only allowed if they have a corresponding Horizontal Pod Autoscaler (HPA) configured in the same namespace. The policy checks for the presence of an HPA that targets the Deployment and denies the creation or update of the Deployment if no such HPA exists. This policy helps enforce scaling practices and ensures that resources are managed efficiently.

## Checklist


other/check-hpa-exists/check-hpa-exists.yaml
```diff
@@ -0,0 +1,44 @@
...
+metadata:
+  name: check-hpa-exists
+  annotations:
+    policies.kyverno.io/title: Ensure HPA for Deployments
+    policies.kyverno.io/category: Other
+    policies.kyverno.io/severity: medium
+    kyverno.io/kyverno-version: 1.11.0
+    policies.kyverno.io/minversion: 1.9.0
+    kyverno.io/kubernetes-version: "1.28"
+    policies.kyverno.io/subject: Deployment
```
chipzoller, line 12: Please set subject annotation to all matching kinds.

PR #1084, opened 2024-07-25
Add policies to enforce or add Istio ambient mode

StevenSmiley:
## Description

Add two polices related to Istio ambient mode: enforce and add.

Ambient mode is an alternative to sidecar mode, for which Kyverno already has policies to enforce or add. Ambient mode is applied using a different namespace label, so these policies enforce or add that label.

## Checklist


JimBugwadia:
Thanks for the contribution @StevenSmiley!

PR #1087, opened 2024-07-27
fix: add separate workflow for CEL policies to unblock e2e tests

Chandan-DK:
## Description
This PR resolves `Strategy expansion exceeded 256 results for job 'chainsaw'` error that is currently blocking the E2E tests from running. 

A separate workflow file has been created for CEL policies and this PR makes use of composite actions to reuse steps in both the workflow files (`test.yml` and `cel-test.yml`)

## Checklist


Chandan-DK:
Made this PR by mistake. But if it

chipzoller:
Yes, I think this would be a good approach! Please do reopen!

realshuting:
Hi @Chandan-DK - let us know if you are still interested in contribution!

realshuting:
Thanks @Chandan-DK !

All new tests failed due to this:
>        === ERROR
        admission webhook "validate-policy.kyverno.svc" denied the request: no unique match for kind GitRepository
    l.go:53: | 18:25:20 | verify-git-repositories | 01 - Create policy and verify  | CREATE    | ERROR | kyverno.io/v1/ClusterPolicy @ verify-git-repositories
        === ERROR
        admission webhook "validate-policy.kyverno.svc" denied the request: no unique match for kind GitRepository

Looks like something is missing, can you take a look?

.github/actions/setup-env/action.yaml
```diff
@@ -0,0 +1,51 @@
...
+    - name: Wait for kyverno ready
+      shell: bash
+      run: |
+        set -e
+        kubectl wait --namespace kyverno --for=condition=ready pod --selector '!job-name' --timeout=60s
+    - name: Install CRDs
+      shell: bash
+      run: |
+        set -e
+        kubectl apply -f ./.chainsaw/crds
```
chipzoller, line 51: This has to point to the CRDs folder for all tests.

chipzoller (changes requested):

.github/actions/setup-env/action.yaml
Chandan-DK: It is pointing to [this](https://github.com/kyverno/policies/tree/main/.chainsaw/crds) folder. We can see that the CRDs are being created when running this command. https://github.com/kyverno/policies/actions/runs/10163856510/job/28107824800#step:3:363

Chandan-DK:
> Thanks @Chandan-DK !
> 
> All new tests failed due to this:
> 
> > ```
> >    === ERROR
> > ```
> 
> ```
>     admission webhook "validate-policy.kyverno.svc" denied the request: no unique match for kind GitRepository
> l.go:53: ���������| 18:25:20 | verify-git-repositories | 01 - Create policy and verify  | CREATE    | ERROR | kyverno.io/v1/ClusterPolicy @ verify-git-repositories
>     === ERROR
>     admission webhook "validate-policy.kyverno.svc" denied the request: no unique match for kind GitRepository
> ```
> 
> Looks like something is missing, can you take a look?

`verify-git-repositories` cel policy is not being created due to this issue (https://github.com/kyverno/kyverno/issues/10313). I have removed this policy for now.

.github/actions/setup-env/action.yaml
chipzoller: Ok, we can reintroduce later.

chipzoller (commented):
Could we rename the `.chainsaw-test` directory of the problematic policy rather than just deleting it?

Chandan-DK:
Sure, but there will still be noise in the CI since Chainsaw executes the `chainsaw-test.yaml` file. We could rename this file too, to prevent noise. What do you think?

chipzoller:
I would ask @eddycharly what's the most reliable thing to do here. I'm not sure what the current recommended approach is for Chainsaw given its present state.

Chandan-DK:
Chainsaw by default looks for the `chainsaw-test.yaml` file inside every folder. We could rename the `.chainsaw-test` directory of the problematic policy to something like `.chainsaw-test-rename-after-issue-10313-fix` to make it easy to identify this problematic policy in the future. Changing the directory name won't stop the `chainsaw-test.yaml` file that is present in it from being executed. So, we should probably name the file this way `.chainsaw-test-rename-after-issue-10313-fix.yaml` and avoid a failing chainsaw test for the time being.

By the way, the policies that have been deleted due to this issue are being tracked [here](https://github.com/kyverno/kyverno/issues/10313). I have been referencing the link of the issue in my commits when deleting policies. We could use those commits to help add the policies back when the issue is resolved.
image

chipzoller:
Ok, that sounds good. Let's give it a try.

chipzoller (approved):
Much thanks! I'm going to go ahead and merge this so we can run our e2e tests again.

realshuting:
Good job @Chandan-DK !

Chandan-DK:
Thanks @realshuting @chipzoller !

PR #1114, opened 2024-08-07
Krishna/deny sa

fast-n-curious:
## Description

For an enhnaced security posture, it is recommended to use specific service accounts and not the default service accounts. These service accounts provide an identity for processes that run in individual Pods and map them to a ServiceAccount object. This policy flags the Pods that use any default service accounts.

## Checklist


fast-n-curious:
Removing this PR as this has both the commit, will resubmit the PR

PR #1115, opened 2024-08-07
Added a new policy: deny-force-delete

fast-n-curious:
## Description

When deleting any resources (Pods/ Deployments/ Services/ Cronjobs etc.), the default timer is 30 seconds which allows a graceful shutdown of the resource. The flag '--grace-period=0 --force' removes this timer and shutsdown the resource immediately. Using this flag may result in data loss as resources as the state may not be saved or lead to potential service disruption. This policy prevents any non-admins to forcefully delete any resources immediately.

## Checklist

### DIFF
--- a/other/deny-force-delete/.chainsaw-test/badpod.yaml
+++ b/other/deny-force-delete/.chainsaw-test/badpod.yaml
@@ -1,0 +1,26 @@
+apiVersion: v1
+kind: Pod
+metadata:
+  labels:
+    run: test1-bad
+  name: test1-bad
+spec:
+  containers:
+  - image: nginx
+    name: test1-bad
+options:
+  gracePeriodSeconds: 0
+---
+# apiVersion: v1
+# kind: Pod
+# metadata:
+#   labels:
+#     run: test2-bad
+#   name: test2-bad
+# spec:
+#   containers:
+#   - image: nginx
+#     name: test2-bad
+# options:
+#   gracePeriodSeconds: 1
+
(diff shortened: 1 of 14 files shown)


other/deny-force-delete/.chainsaw-test/goodpod.yaml
chipzoller, line 1: `options` is an invalid field for Pods and all other resources. This field will only be populated in the AdmissionReview if a DELETE request is sent with a flag such as `--grace-period`. You will have to use a command/script step to perform a deletion using that flag.

other/deny-force-delete/.chainsaw-test/badpod.yaml
chipzoller, line 1: See comments on `goodpod.yaml`.

other/deny-force-delete/.kyverno-test/resource.yaml
chipzoller, line 1: See comments on Chainsaw test resources. The `options` field is invalid in an actual resource spec. For the Kyverno CLI, you will need to set a variable instead which looks like you're doing. All resources used in testing must be valid resources.

chipzoller (changes requested):

fast-n-curious:
@chipzoller not sure why it's not able to verify the signature. As can be seen from the below screenshot, I have reverted my previous changes and submitted them with signature again. It's still throwing an error!

Screenshot 2024-08-14 at 3 00 04 PM

chipzoller:
You need to sign all your commits per DCO. Please also resolve all conversations when you push changes which address them.

fast-n-curious:
Will open a new PR with the proper signature for DCO to pass.

PR #1121, opened 2024-08-09
feat(ps): add mutate pod security policies

KhaledEmaraDev:
## Description

This adds Mutate Pod Security Policies for `allowPrivilegeEscalation` and `runAsNonRoot`.

## Checklist


pod-security/mutate/set-run-as-non-root/set-run-as-non-root.yaml
```diff
@@ -0,0 +1,87 @@
+apiVersion: kyverno.io/v1
+kind: ClusterPolicy
+metadata:
+  name: set-run-as-non-root
+  annotations:
+    policies.kyverno.io/title: Set runAsNonRoot
+    policies.kyverno.io/category: Pod Security Standards (Mutate)
+    policies.kyverno.io/severity: medium
+    policies.kyverno.io/subject: Pod
+    kyverno.io/kyverno-version: 1.6.0
```
chipzoller, line 10: This label reflects the version of Kyverno on which the policy was tested and confirmed working. Should be the 1.12 release.

pod-security/mutate/set-run-as-non-root/set-run-as-non-root.yaml
```diff
@@ -0,0 +1,87 @@
...
+    kyverno.io/kubernetes-version: "1.22-1.30"
```
chipzoller, line 11: Have you personally verified this works across that many versions? Again, this annotation holds what was tested not what "should" work.

pod-security/mutate/set-run-as-non-root/set-run-as-non-root.yaml
```diff
@@ -0,0 +1,87 @@
...
+    policies.kyverno.io/description: >-
+      Containers must be required to run as non-root users. This policy ensures
+      `runAsNonRoot` is set to `true`.
```
chipzoller, line 14: Description is not accurate.

pod-security/mutate/set-run-as-non-root/set-run-as-non-root.yaml
```diff
@@ -0,0 +1,87 @@
...
+      match:
+        all:
+          - resources:
+              kinds:
+                - Pod
+      preconditions:
+        all:
+          - key: "{{ type(request.object.spec.containers) }}"
+            operator: Equals
+            value: "array"
```
chipzoller, line 29: This is unnecessary. `containers[]` is a required object for a Pod and is always an array.

pod-security/mutate/set-run-as-non-root/set-run-as-non-root.yaml
chipzoller, line 1: All these rules can be condensed into just one with multiple `foreach` loops. We have several examples of this already.

chipzoller (changes requested):
Haven't done a complete review here but this is a start. I'm also wondering if there's a better category and folder location for these.

pod-security/mutate/set-run-as-non-root/set-run-as-non-root.yaml
KhaledEmaraDev: @chipzoller  I started with that. But it's hard to do with JSON Patch unlike PSM. We need to put the  container type in the path and check existence.

Important Context. We are trying to move people to JSON Patch because PSM performance is not ideal right now.

pod-security/mutate/set-run-as-non-root/set-run-as-non-root.yaml
chipzoller: Yes, in that case you would.

Issue #1166, opened 2024-09-24
[Bug] Security Capabilites must omit the CAP prefix which is required by Kyverno policies


StraysWonderland:
### Kyverno Version

1.11

### Kubernetes Version

1.26

### Kubernetes Platform

EKS

### Description

We have implemented a subset of the best practices policies on our aks cluster.
The issue I'm facing is related to the require-drop-cap-net-raw policy which can be found here [within the kyverno policies repository](https://github.com/kyverno/policies/blob/main/best-practices/require-drop-cap-net-raw/require-drop-cap-net-raw.yaml)

However this policy does only recognize a pod as passing the policy if the capability section is defined as:
```yaml
securityContext:
  capabilities:
    drop:
      - CAP_NET_RAW
```
so with the CAP prefix

According to the official[ kubernetes documentation](https://kubernetes.io/docs/tasks/configure-pod-container/security-context/#set-capabilities-for-a-container) however, these capabilities should omit the CAP prefix as can be seen here:
![image](https://github.com/user-attachments/assets/71114079-bb30-406e-a113-cc5fbf7af1da)

And the problem is that kyverno does not seem to recognize resources dropping the capabilities in this manner:
```yaml
securityContext:
  capabilities:
    drop: ["NET_RAW"]
```

Ive done a bit of testing on some pods and it seems that pods with the CAP_NET_RAW definition, tough passing the kyverno policy, are not actually dropping the security capabilities while pods with the NET_RAW definition fail kyverno but are correctly dropping the security capability.

[...]

PR #1168, opened 2024-09-26
Fix for issue 1166: Allow CAP_NET_RAW and NET_RAW drop capabilities

epasham:
## Related Issue(s)

kyverno does not seem to recognize resources dropping the capabilities as given below

securityContext:
  capabilities:
    drop: ["NET_RAW"]

## Description

require-cap-net-raw policy only allows securityContext capabilities drop for CAP_NET_RAW.
Good option is to allow both CAP_NET_RAW and NET_RAW drop capability.

closes https://github.com/kyverno/policies/issues/1166
## Checklist


realshuting:
@epasham - can you add "closes #1166" in the PR description to automatically close the issue?

best-practices/require-drop-cap-net-raw/require-drop-cap-net-raw.yaml
```diff
@@ -25,19 +26,19 @@ spec:
...
+            - name: mustDropCapabilities
+              expression: "['CAP_NET_RAW','NET_RAW']"
+            - name: allContainers
+              expression: "object.spec.containers + object.spec.?initContainers.orValue([]) + object.spec.?ephemeralContainers.orValue([])"
+          expressions:
+            - expression: >-
+                variables.allContainers.all(container, 
+                container.?securityContext.?capabilities.?drop.orValue([]).exists(capability, capability.upperAscii() in variables.mustDropCapabilities))
+              message: >-
+                Containers must drop the `CAP_NET_RAW` capability.
```
MariamFahmy98, line 44: Why do we use the `validate.cel` subrule here? We already have the best practices written in CEL expressions.

I expect to solve the bug using the existing foreach.

epasham: I picked up the wrong policy file. I should have picked following one -> https://github.com/kyverno/policies/blob/main/best-practices-cel/require-drop-cap-net-raw/require-drop-cap-net-raw.yaml
will revert the changes and update the correct policy file

PR #1179, opened 2024-10-07
fix disallow-privilege-escalation policy

JimBugwadia:
## Related Issue(s)

n/a

## Description

The pod security standard restricted mode does not allow a `nil` value for `allowPrivilegeEscalation`. 

It must be set to `false`.

```
    securityContext:
      allowPrivilegeEscalation: false
```

## Checklist

### DIFF
--- a/pod-security-cel/restricted/disallow-privilege-escalation/disallow-privilege-escalation.yaml
+++ b/pod-security-cel/restricted/disallow-privilege-escalation/disallow-privilege-escalation.yaml
@@ -37,7 +37,7 @@ spec:
           expressions:
             - expression: >- 
                 variables.allContainers.all(container, 
-                container.?securityContext.?allowPrivilegeEscalation.orValue(true) == false)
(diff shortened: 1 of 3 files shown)


pod-security-cel/restricted/disallow-privilege-escalation/disallow-privilege-escalation.yaml
```diff
@@ -37,7 +37,7 @@ spec:
           expressions:
             - expression: >- 
                 variables.allContainers.all(container, 
-                container.?securityContext.?allowPrivilegeEscalation.orValue(false) == false)
+                container.?securityContext.?allowPrivilegeEscalation.orValue(true) == false)
```
eddycharly, line 40:
```suggestion
                container.?securityContext.allowPrivilegeEscalation.orValue(true) == false)
```

eddycharly (approved):

PR #1237, opened 2025-02-17
fix(tests): result != expected should fail a test

bakunowski:
## Related Issue(s)

This is required for https://github.com/kyverno/kyverno/pull/12117 to be merged.

## Description

This PR fixes two tests that rely on wrong Kyverno behaviour.

In the case of `allowed-label-changes`:
- test updating a non `breakglass` label should fail, but it succeeds. Prior to https://github.com/kyverno/kyverno/pull/12117 this is considered a `Pass` result.

In the case of `prepend-image-registry`:
- I'm not entirely sure what the fail case for a mutating policy is, apart from the webhook being unavailable. We could probably fake it with a `resource.operation` set to something other than `CREATE` or `UPDATE`, but I've decided to delete the `result: fail` tests, and leave the tests checking if mutation happened correctly.

## Checklist

- [x] I have added test manifests and resources covering both positive and negative tests that prove this policy works as intended. (with the caveat of the mutating policy)


JimBugwadia:
@bakunowski - can you please help fix the conflict?

PR #1251, opened 2025-03-08
adding cosign signed image policy to maintain supply chain security in Tekton

arunvel1988:
## Checklist


arunvel1988:
not solved

PR #1255, opened 2025-03-15
Update cleanup-bare-pods.yaml

suhash-b:
## Checklist


younsl:
According to the [Kyverno documentation on cleanup policies](https://kyverno.io/docs/writing-policies/cleanup/), in the latest version (v1.13.x), the `kind` must remain ClusterCleanupPolicy. ClusterPolicy does not support cleanup policies.

JimBugwadia (changes requested):
@suhash-b - can you explain what you are trying to change? 

`ClusterCleanupPolicy` seems to be the right kind for this policy.

PR #1262, opened 2025-03-21
other-vpol

Mohdcode:
sample policy conversion for vpol

### DIFF
--- a/other-vpol/allowed-annotations/.chainsaw-test/chainsaw-test.yaml
+++ b/other-vpol/allowed-annotations/.chainsaw-test/chainsaw-test.yaml
@@ -1,0 +1,44 @@
+# yaml-language-server: $schema=https://raw.githubusercontent.com/kyverno/chainsaw/main/.schemas/json/test-chainsaw-v1alpha1.json
+apiVersion: chainsaw.kyverno.io/v1alpha1
+kind: Test
+metadata:
+  creationTimestamp: null
+  name: allowed-annotations
+spec:
+  # disable templating because it can cause issues with ValidatingPolicy
+  template: false
+  steps:
+  - name: step-01
+    try:
+    - apply:
+        file: ../allowed-annotations.yaml
+    - patch:
+        resource:
+          apiVersion: policies.kyverno.io/v1alpha1
+          kind: ValidatingPolicy
+          metadata:
+            name: allowed-annotations
+          spec:
+            validationActions:
+             - Deny
+    - assert:
+        file: policy-ready.yaml
+    - sleep: 
+       duration: 10s
+  - name: step-02
+    try:
+    - apply:
+        file: pod-good.yaml
+    - apply:
+        file: podcontroller-good.yaml
+    - apply:
+        expect:
+        - check:
+            ($error != null): true
+        file: pod-bad.yaml
+    - apply:
+        expect:
+        - check:
+            ($error != null): true
+        file: podcontroller-bad.yaml
+
--- a/other-vpol/allowed-annotations/.chainsaw-test/pod-bad.yaml
+++ b/other-vpol/allowed-annotations/.chainsaw-test/pod-bad.yaml
@@ -1,0 +1,47 @@
+apiVersion: v1
+kind: Pod
+metadata:
+  annotations:
+    fluxcd.io/cat: meow
+  name: badpod01
+spec:
+  containers:
+    - name: pod01-01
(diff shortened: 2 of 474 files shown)


other-vpol/advanced-restrict-image-registries/advanced-restrict-image-registries.yaml
```diff
@@ -0,0 +1,46 @@
...
+  matchConstraints:
+    resourceRules:
+    - apiGroups:   [""]
+      apiVersions: ["v1"]
+      operations:  ["CREATE", "UPDATE"]
+      resources:   ["pods"]
+  variables:
+            - name: cm
+              expression: >-
+               resource.Get("v1", "configmaps", "default", "clusterregistries")
```
JimBugwadia, line 35: How will this handle a list of registries in the ConfigMap? Same for the annotation?

JimBugwadia (changes requested):

other-vpol/advanced-restrict-image-registries/advanced-restrict-image-registries.yaml
Mohdcode: Updated for list handling: it can process lists of registries and annotations. However, multiple annotation keys will require separate handling.
```yaml
- expression: >-
        variables.allContainers.all(container, 
         variables.nsregistries.contains(container.image.split('/')[0] + '/') || 
          variables.registriesData.contains(container.image.split('/')[0] + '/')
        )
```

other-vpol/advanced-restrict-image-registries/advanced-restrict-image-registries.yaml
```diff
@@ -30,17 +27,21 @@ spec:
...
+    - name: registriesData
+      expression: >-
+        variables.cm.data[?'registries'].orValue('')
   validations:
-            - expression: "variables.allContainers.all(container, container.image.startsWith(variables.nsregistries) || container.image.startsWith(variables.registriesData))"
-              message: This Pod names an image that is not from an approved registry.
-
+    - expression: >-
+        variables.allContainers.all(container, 
+         variables.nsregistries.contains(container.image.split('/')[0] + '/') ||
```
JimBugwadia, line 44: Why split by `/`?

I suggest something like this:


```yaml
apiVersion: v1
kind: ConfigMap
metadata:
  name: allowed-registries
  namespace: platform
data:
  registries: "foo.com, bar.io"

```


```yaml

apiVersion: policies.kyverno.io/v1alpha1
kind: ValidatingPolicy
metadata:
  name: restrict-image-registries
spec:
  validationActions: 
    - Deny
  evaluation:
    background:
      enabled: true
  matchConstraints:
    resourceRules:
    - apiGroups:   [""]
      apiVersions: ["v1"]
      operations:  ["CREATE", "UPDATE"]
      resources:   ["pods"]
  variables:
  - name: registriesConfigMap
    expression: >-
      resource.Get("v1", "configmaps", "platform", "allowed-registries")
  - name: allowedRegistries
    expression: "variables.registriesConfigMap.data['registries'].split(',')"
  - name: allContainers
    expression: "object.spec.containers 
      + object.spec.?initContainers.orValue([]) 
      + object.spec.?ephemeralContainers.orValue([])"
  validations:
  - message: This Pod has an image that is not from an approved registry.
    expression: "variables.allContainers.all(container, 
      variables.allowedRegistries.exists(allowedRegistry,
        container.image.startsWith(allowedRegistry.trim())))"
```

JimBugwadia (changes requested):

other-vpol/advanced-restrict-image-registries/advanced-restrict-image-registries.yaml
Mohdcode: I was experimenting with three formats for defining registries in the `ConfigMap:` a `YAML` list with hyphens `(- ghcr.io/)`, a `JSON` array `(["img.corp.com/", "docker.io/"])` , and a newline-separated string `(ghcr.io/\nmyregistry/)`. At that time, I hadn’t tested using values like `"docker.io, ghcr.io"` without trailing slashes, otherwise I might have approached the validation differently. Thanks for the solution as well

JimBugwadia: @Mohdcode - I recommend removing the need for users to specify a "/" and simply allowing a comma separated list, which is fairly typical and easy to use.

PR #1345, opened 2025-09-25
Update require-pod-requests-limits.yaml

22030083:
## Checklist


fjogeleit:
The lowercase value `audit` is deprecated, so there is no reason to change it.

PR #1349, opened 2025-09-30
Converting ClusterPolicy to MutatingPolicy [patch 1]

Mohab96:
## Related Issue(s)

https://github.com/kyverno/kyverno/issues/13709

## Description

This PR is the second patch to convert the traditional `ClusterPolicy` to the new v1alpha1 `MutatingPolicy`.
This here is 10 policies, and later patches will be 10 policies per PR.

## Checklist


JimBugwadia:
@Mohab96 - are we converting existing policies or adding new ones?

We will continue to support ClusterPolicy mutate rules for some period of time, so will be good to have both until we are ready to deprecate ClusterPolicy.

Issue #1488, opened 2026-05-13
[Sample] Convert old verify image ClusterPolicy to IVPOL


naseyro:
### Problem Statement

Convert old SLSA provenance ClusterPolicy to the new ImageValidating policies

### Solution Description

Convert old SLSA provenance ClusterPolicy to the new ImageValidating policies

currently working on: 
- [SLSA provenance](https://github.com/kyverno/policies/pull/1487)
- [verify-image-with-multi-keys](https://github.com/kyverno/policies/pull/1489)

### Troubleshooting

- [x] I have read and followed the documentation AND the [troubleshooting guide](https://kyverno.io/docs/troubleshooting/).
- [x] I have verified the policy does not exist in the [samples library](https://kyverno.io/policies/).

PR #1489, opened 2026-05-14
Convert old verify-image-with-multi-keys to ivpol

naseyro:
## Related Issue(s)

Part of #1488 

Old ClusterPolicy: https://github.com/kyverno/policies/tree/main/other/verify-image-with-multi-keys

## Description

Convert `verify-image-with-multi-keys` to the new IVPOL

## Checklist

### DIFF
--- a/other-ivpol/verify-image-with-multi-keys/.chainsaw-test/bad.yaml
+++ b/other-ivpol/verify-image-with-multi-keys/.chainsaw-test/bad.yaml
@@ -1,0 +1,10 @@
+apiVersion: v1
+kind: Pod
+metadata:
+  name: bad-pod
+  namespace: test-signing
+spec:
+  containers:
+    - name: test-container
+      image: 'ghcr.io/kyverno/test-verify-image:unsigned'
+      imagePullPolicy: Always
--- a/other-ivpol/verify-image-with-multi-keys/.chainsaw-test/chainsaw-test.yaml
+++ b/other-ivpol/verify-image-with-multi-keys/.chainsaw-test/chainsaw-test.yaml
@@ -1,0 +1,32 @@
+# yaml-language-server: $schema=https://raw.githubusercontent.com/kyverno/chainsaw/main/.schemas/json/test-chainsaw-v1alpha1.json
+apiVersion: chainsaw.kyverno.io/v1alpha1
+kind: Test
+metadata:
+  creationTimestamp: null
+  name: verify-image-with-multi-keys
+spec:
+  # disable templating because it can cause issues with CEL expressions
+  template: false
+
+  steps:
+  - name: step-01
+    try:
+    - script:
+        content: kubectl create namespace test-signing || true
+    - apply:
+        file: ../verify-image-with-multi-keys.yaml
+    - apply:
+        file: configmap.yaml
+    - assert:
+        file: policy-ready.yaml
+  - name: step-02
+    try:
+    - create:
+        file: good.yaml
+        timeout: 360s
+    - create:
+        timeout: 360s
+        expect:
+        - check:
+            ($error != null): true
+        file: bad.yaml
--- a/other-ivpol/verify-image-with-multi-keys/.chainsaw-test/good.yaml
+++ b/other-ivpol/verify-image-with-multi-keys/.chainsaw-test/good.yaml
@@ -1,0 +1,10 @@
+apiVersion: v1
+kind: Pod
+metadata:
+  name: good-pod
+  namespace: test-signing
+spec:
+  containers:
+    - name: test-container
+      image: 'ghcr.io/kyverno/test-verify-image@sha256:b588b277f929d067df06f8a88edfc4c5f9b0ec4349aa426e74e2923df46b3656'
+      imagePullPolicy: Always
--- a/other-ivpol/verify-image-with-multi-keys/artifacthub-pkg.yml
+++ b/other-ivpol/verify-image-with-multi-keys/artifacthub-pkg.yml
@@ -1,0 +1,23 @@
+name: verify-image-with-multi-keys-ivpol
+version: 1.0.0
+displayName: Verify Image with Multiple Keys (IVPOL)
+description: >-
+  There may be multiple keys used to sign images based on the parties involved in the creation process. This image verification policy requires the named image be signed by two separate keys.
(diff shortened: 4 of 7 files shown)


other-ivpol/verify-image-with-multi-keys/verify-image-with-multi-keys.yaml
```diff
@@ -0,0 +1,50 @@
...
+        resources: ["pods"]
+  matchImageReferences:
+    - glob: "ghcr.io/myorg/myimage*"
+  attestors:
+    - name: productionKey
+      cosign:
+        key:
+          data: |-
+            -----BEGIN PUBLIC KEY-----
+            Replace with your production public key in PEM format.
```
realshuting: How does this policy work without a key?

realshuting (changes requested):

other-ivpol/verify-image-with-multi-keys/verify-image-with-multi-keys.yaml
naseyro: @realshuting it's just a placeholder, do I have to provide a key?

realshuting: yes

other-ivpol/verify-image-with-multi-keys/.chainsaw-test/bad.yaml
```diff
@@ -0,0 +1,9 @@
+apiVersion: v1
+kind: Pod
+metadata:
+  name: bad-pod
+spec:
+  containers:
+    - name: test-container
+      image: 'ghcr.io/myorg/myimage:v1'
```
lucchmielowski: If we provide a key, wondering whether we should also provide a proper image for the chainsaw test ? (since the test shouldn't fail because of a `ImagePullBackoff` 🤔

naseyro: I was adding it as just a placeholder, but is there a specific image I should use here? @lucchmielowski

lucchmielowski: I was thinking about using this kind of image: 
https://github.com/kyverno/kyverno/blob/1350d6a5e353c43444adf92c1c02191eb282221a/test/conformance/chainsaw/image-validating-policies/match-conditions/bad-pod.yaml#L11

other-ivpol/verify-image-with-multi-keys/.chainsaw-test/good.yaml
```diff
@@ -0,0 +1,9 @@
+apiVersion: v1
+kind: Pod
+metadata:
+  name: good-pod
+spec:
+  containers:
+    - name: test-container
+      image: 'nginx:latest'
```
naseyro: @lucchmielowski  I see this one used somewhere here, should I use it instead of nginx:latest?

https://github.com/kyverno/policies/blob/76be98a25d49ae01278a94ecde8f50f9e08577ef/other/verify-image-ivpol/.chainsaw-test/good.yaml#L8

other-ivpol/verify-image-with-multi-keys/verify-image-with-multi-keys.yaml
```diff
@@ -0,0 +1,52 @@
...
+    background:
+      enabled: false
+  matchConstraints:
+    resourceRules:
+      - apiGroups: [""]
+        apiVersions: ["v1"]
+        operations: ["CREATE", "UPDATE"]
+        resources: ["pods"]
+  matchImageReferences:
+    - glob: "ghcr.io/myorg/myimage*"
```
naseyro:
```suggestion
    - glob: ghcr.io/kyverno/test-verify-image*
```

other-ivpol/verify-image-with-multi-keys/verify-image-with-multi-keys.yaml
```diff
@@ -0,0 +1,56 @@
...
+  - name: namespaceKey
+    cosign:
+      key:
+        expression: "variables.keys.data[object.metadata.namespace]"
+      rekor:
+        ignoreTlog: true
+  validations:
+  - message: "Image is missing a valid signature from the production key."
+    expression: >-
+      images.containers.all(image,
```
Suhani95, line 50: The validation currently checks only `images.containers`. Kyverno extracts Pod images into `containers`, `initContainers`, and `ephemeralContainers`, so matching images in init containers can currently bypass verification because `all()` on an empty `images.containers` list evaluates to true. Could we extend validation coverage to the other Pod image buckets as well?

other-ivpol/verify-image-with-multi-keys/.chainsaw-test/chainsaw-test.yaml
```diff
@@ -0,0 +1,32 @@
...
+    - apply:
+        file: ../verify-image-with-multi-keys.yaml
+    - apply:
+        file: configmap.yaml
+    - assert:
+        file: policy-ready.yaml
+  - name: step-02
+    try:
+    - create:
+        file: good.yaml
```
Suhani95, line 25: Could we also add Chainsaw coverage for the init container path? That would help prevent regressions around image bucket handling in future IVPOL migrations.

other-ivpol/verify-image-with-multi-keys/artifacthub-pkg.yml
```diff
@@ -0,0 +1,24 @@
...
+install: |-
+  ```shell
+  kubectl apply -f https://raw.githubusercontent.com/kyverno/policies/main/other-ivpol/verify-image-with-multi-keys/verify-image-with-multi-keys.yaml
+  ```
+keywords:
+  - kyverno
+  - Software Supply Chain Security
+
+readme: |
+  There may be multiple keys used to sign images based on the parties involved in the creation process. This image verification policy requires the named image be signed by two separate keys. Replace the placeholder public keys with your own production and namespace-specific keys.
```
Suhani95: Could we clarify that users must create the `default/keys` ConfigMap with both production and namespace-specific entries? The policy currently depends on that ConfigMap, but the readme does not explicitly mention it.

PR #1491, opened 2026-05-15
Add label-node-selector policy

kappa8219:
Actually the idea is to use field "role" in K8s UI-s for marking (the cattle) nodes.

This policy adds "well known" label node-role.kubernetes.io/ROLE= to each node. It is impossible to do on a bootstrap phase so the idea is to do it later with Kyverno. 

It looks like(FreeLens):
image

K9s:
image

So in general it is easier to observe and operate those marked nodes.

P.S. Includes Artifact Hub metadata, a Chainsaw test, and RBAC notes for the Kyverno background controller and Node resource filter.


kappa8219:
PTAL @realshuting @MariamFahmy98 
All required files are in place and DCO is passing. Thanks for reviewing!

PR #1504, opened 2026-07-03
Fix RCE bypass in restrict-clusterrole-nodesproxy policies (Fixes #1492)

### LINKED ISSUE #1492: [Bug] restrict-clusterrole-nodesproxy does not block apiGroups wildcard "*" — RCE bypass
## Kyverno Version

1.13 (tested on v1.18.0)

## Kubernetes Version

1.35

## Kubernetes Platform

Bare metal

## Description

The `restrict-clusterrole-nodesproxy` policy (both CEL and JMESPath variants) does not block a `ClusterRole` that uses `apiGroups: ["*"]` with an explicit `resources: ["nodes/proxy"]`. Yet Kubernetes treats `"*"` as a wildcard matching all API groups including the core group `""`, so such a ClusterRole grants the exact same permissions as `apiGroups: [""]` — which the policy correctly blocks.

This is a regression introduced by PR #931, which tightened the policy to only match `apiGroup == ""` in order to fix the over-blocking reported in #895 (Istio-style ClusterRoles with `resources: ["*"]` in custom API groups). That fix was correct, but it inadvertently left a gap: a ClusterRole with `nodes/proxy` **explicitly** listed in resources combined with `apiGroups: ["*"]` now bypasses the policy entirely.

This gap is distinct from the #895 case. In #895, `nodes/proxy` was never explicitly requested — it was implied by a resource wildcard in unrelated API groups. Here, `nodes/proxy` is explicitly requested, and the apiGroups wildcard is genuinely dangerous.

The bypass is fully exploitable: a ServiceAccount bound to such a ClusterRole can execute arbitrary commands in any Pod on reachable Nodes via WebSocket to the Kubelet API port 10250 (see [Graham Helton's research](https://grahamhelton.com/blog/nodes-proxy-rce)), bypassing `CREATE` verb checks and without generating `pods/exec` audit logs.

## Steps to reproduce

**1. Deploy the policy in Enforce mode.**

**2. Apply a ClusterRole with `apiGroups: [""]` — correctly blocked:**

```
$ kubectl apply -f clusterrole-empty-apigroup.yaml
Error from server: admission webhook "validate.kyverno.svc-fail" denied the request:
resource ClusterRole//test-nodesproxy-empty-apigroup was blocked due to the following policies
restrict-clusterrole-nodesproxy: A ClusterRole containing nodes/proxy is not allowed.
```

**3. Apply a ClusterRole with `apiGroups: ["*"]` — bypasses the policy:**

```
$ kubectl apply -f clusterrole-wildcard-apigroup.yaml
clusterrole.rbac.authorization.k8s.io/test-nodesproxy-wildcard-apigroup created
```

**4. Confirm the bound ServiceAccount has GET but not CREATE:**

```
$ kubectl auth can-i get nodes --subresource=proxy --as=system:serviceaccount:test-ns:test-sa
yes
$ kubectl auth can-i create nodes --subresource=proxy --as=system:serviceaccount:test-ns:test-sa
no
```

**5. POST to kubelet `/exec` is correctly denied:**

```
$ curl -sk -X POST -H "Authorization: Bearer $TOKEN" \
    "https://$NODE_IP:10250/exec/$NS/$POD/$CONTAINER?command=id&stdout=true"
Forbidden (user=system:serviceaccount:test-ns:test-sa, verb=create, resource=nodes, subresource(s)=[proxy])
```

**6. WebSocket to kubelet `/exec` succeeds — RCE:**

```
# Python websockets library, connecting directly to kubelet port 10250
# using the token of the ServiceAccount bound to the wildcard ClusterRole

uid=999(redis) gid=1000(redis) groups=1000(redis)
{"metadata":{},"status":"Success"}
```

The Kubelet authorizes WebSocket connections based on the HTTP `GET` required by the WebSocket handshake (RFC 6455), mapping it to the RBAC `GET` verb. It never performs a secondary check for `CREATE`. So `nodes/proxy GET` is sufficient for command execution in any Pod. This is documented in detail by Graham Helton: https://grahamhelton.com/blog/nodes-proxy-rce

## Expected behavior

A `ClusterRole` with `apiGroups: ["*"]` and `resources: ["nodes/proxy"]` should be blocked, since `"*"` matches all API groups including `""` and this combination is fully exploitable.

The fix should **not** regress the #895 case: Istio-style ClusterRoles with `resources: ["*"]` in custom API groups (without explicit `nodes/proxy`) should continue to pass, and they do — the proposed fix only adds a check on `apiGroups`, not on `resources`.

**Proposed fix for the CEL expression:**

```
object.rules == null ||
!object.rules.exists(rule,
  has(rule.resources) &&
  rule.resources.exists(resource, resource == 'nodes/proxy') &&
  rule.apiGroups.exists(apiGroup, apiGroup == '' || apiGroup == '*')
)
```

The same fix should be applied to the JMESPath variant.

**Additional test case to add:**

```yaml
# Should FAIL — explicit nodes/proxy with wildcard apiGroup
apiVersion: rbac.authorization.k8s.io/v1
kind: ClusterRole
metadata:
  name: clusterrole-nodesproxy-wildcard-apigroup
rules:
  - apiGroups: ["*"]
    resources: ["nodes/proxy"]
    verbs: ["get"]
```

I can draft a PR to fix both CEL and legacy versions of the "Restrict ClusterRole with Nodes Proxy" policies.

## Screenshots

_N/A — textual output provided above._

## Kyverno logs

_No relevant logs — the issue is a missing match condition, not an error._

## Slack discussion

_None._

dolliecoder:
## Related Issue(s)

Fixes #1492

## Description

This PR fixes a Remote Code Execution (RCE) bypass vulnerability in the `restrict-clusterrole-nodesproxy` policy, as detailed in #1492.

Previously, the policy successfully blocked ClusterRoles requesting access to the `nodes/proxy` resource on the core API group (`apiGroups: [""]`), but failed to block requests utilizing the wildcard API group (`apiGroups: ["*"]`). 

The logic has been updated to explicitly match and deny both `""` and `"*"` across all three variants of the policy:
- Legacy JMESPath (`other`)
- CEL ClusterPolicy (`other-cel`)
- CEL ValidatingPolicy (`other-vpol`)

## Checklist

### DIFF
--- a/other/restrict-clusterrole-nodesproxy/restrict-clusterrole-nodesproxy.yaml
+++ b/other/restrict-clusterrole-nodesproxy/restrict-clusterrole-nodesproxy.yaml
@@ -35,6 +35,10 @@ spec:
             - key: nodes/proxy
               operator: AnyIn
               value: "{{ request.object.rules[].resources[] }}"
+            any:
             - key: ""
+              operator: AnyIn
+              value: "{{ request.object.rules[].apiGroups[] }}"
+            - key: "*"
               operator: AnyIn
               value: "{{ request.object.rules[].apiGroups[] }}"
--- a/other-cel/restrict-clusterrole-nodesproxy/restrict-clusterrole-nodesproxy.yaml
+++ b/other-cel/restrict-clusterrole-nodesproxy/restrict-clusterrole-nodesproxy.yaml
@@ -38,6 +38,6 @@ spec:
                 !object.rules.exists(rule,
                 has(rule.resources) &&
                 rule.resources.exists(resource, resource == 'nodes/proxy') && 
-                rule.apiGroups.exists(apiGroup, apiGroup == ''))
+                rule.apiGroups.exists(apiGroup, apiGroup == '' || apiGroup == '*'))
               message: "A ClusterRole containing the nodes/proxy resource is not allowed."
 
--- a/other-vpol/restrict-clusterrole-nodesproxy/restrict-clusterrole-nodesproxy.yaml
+++ b/other-vpol/restrict-clusterrole-nodesproxy/restrict-clusterrole-nodesproxy.yaml
@@ -34,6 +34,6 @@ spec:
         object.rules == null || 
         !object.rules.exists(rule, 
         rule.resources.exists(resource, resource == 'nodes/proxy') && 
-        rule.apiGroups.exists(apiGroup, apiGroup == ''))
+        rule.apiGroups.exists(apiGroup, apiGroup == '' || apiGroup == '*'))
       message: "A ClusterRole containing the nodes/proxy resource is not allowed."
 
--- a/other/restrict-clusterrole-nodesproxy/.chainsaw-test/cr-bad.yaml
+++ b/other/restrict-clusterrole-nodesproxy/.chainsaw-test/cr-bad.yaml
@@ -18,3 +18,15 @@ rules:
 - apiGroups: [""]
   resources: ["pods", "nodes/proxy"]
   verbs: ["get", "watch", "list"]
+---
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRole
+metadata:
+  name: badcr03
+rules:
+- apiGroups: ["*"]
+  resources: ["nodes/proxy", "namespaces"]
+  verbs: ["get", "watch", "list"]
+- apiGroups: ["apps"]
+  resources: ["deployments"]
+  verbs: ["get", "watch", "list"]


other/restrict-clusterrole-nodesproxy/restrict-clusterrole-nodesproxy.yaml
```diff
@@ -35,6 +35,10 @@ spec:
             - key: nodes/proxy
               operator: AnyIn
               value: "{{ request.object.rules[].resources[] }}"
+            any:
```
Suhani95, line 38: Could we add a regression test for both apiGroups: [""] and apiGroups: ["*"] to ensure this bypass doesn't reappear?

dolliecoder: done

PR #1508, opened 2026-07-11
feat: block system:authenticated in restrict-binding-system-groups

rrskris:
## What this does

`restrict-binding-system-groups` blocks bindings to `system:anonymous`, `system:unauthenticated`, and `system:masters` — but not **`system:authenticated`**, the group containing every principal with a valid token (all users and all ServiceAccounts). Binding a privileged role to it is a classic cluster-takeover misconfiguration (kubectl itself prints a warning when you create one) and is broader in blast radius than the groups the policy already covers.

## Changes

- adds a `restrict-authenticated` rule mirroring the existing three (classic `other/`)
- adds the equivalent CEL expression to the `other-cel/` and `other-vpol/` variants
- updates the policy descriptions to mention the fourth group
- extends the `.kyverno-test` and `.chainsaw-test` resources with failing `system:authenticated` bindings (`badcrb04`/`badrb04`) in all three variants

`kyverno test` passes on all three variants (32 / 14 / 14 results). Test files normalized with `kyverno fix test --save`; artifacthub digests refreshed via `.hack/update-artifacthub-pkg.sh` (idempotent).

### DIFF
--- a/other-cel/restrict-binding-system-groups/.chainsaw-test/crb-bad.yaml
+++ b/other-cel/restrict-binding-system-groups/.chainsaw-test/crb-bad.yaml
@@ -39,3 +39,16 @@ roleRef:
   name: manager
   apiGroup: rbac.authorization.k8s.io
 
+---
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRoleBinding
+metadata:
+  name: badcrb04
+subjects:
+- kind: Group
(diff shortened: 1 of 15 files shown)


rrskris:
Checking in on this one too — happy to rebase or rework the policy if there's a direction you'd prefer. Let me know if adding kuttl tests or an artifacthub entry would help get it over the line.

PR #1522, opened 2026-08-08
feat: Add `other/add-daemonset-tolerations` — new DaemonSet toleration-defaulting policy

yogeshbendre:
## Related Issue(s)

Not a `kyverno/policies` issue — motivated by a pattern found while auditing several unrelated
Kubernetes operators for DaemonSet configuration gaps:

- [NVIDIA/gpu-operator#2717](https://github.com/NVIDIA/gpu-operator/issues/2717)
- [k8snetworkplumbingwg/sriov-network-operator#1132](https://github.com/k8snetworkplumbingwg/sriov-network-operator/issues/1132)

Both are the same shape: a DaemonSet manifest with zero `tolerations` set, silently failing to
schedule on tainted nodes, while every sibling DaemonSet in the same operator hardcodes a
toleration. Filed upstream against each operator individually, but the pattern recurring across
unrelated projects suggested a cluster-level safety net was worth contributing here too.

## Description

Adds `other/add-daemonset-tolerations`: a `mutate` policy that adds a broad `operator: Exists`
toleration to any `DaemonSet` whose Pod template has **no tolerations at all**. Gated by a
JMESPath precondition checking that `length()` of the existing tolerations list (defaulting to an
empty list if the field is absent) equals `0`, so it only fills a true gap — any DaemonSet with
even one toleration already configured is left untouched.

Checked the existing library first: `other/add-tolerations` exists but is `Pod`-scoped with one
hardcoded, organization-specific example toleration (`org.com/role=service:NoSchedule`) — a
customize-me template, not a DaemonSet-aware safety net for the "zero tolerations" case. No
existing policy covers this specific gap.

**Validation**, using the real (sanitized) manifests from the two linked issues plus five
already-correct sibling DaemonSets as negative controls:

```
$ kyverno test other/add-daemonset-tolerations/.kyverno-test
Test Summary: 7 tests passed and 0 tests failed

$ chainsaw test other/add-daemonset-tolerations/.chainsaw-test
--- PASS: chainsaw/add-daemonset-tolerations (5.79s)
```

The `chainsaw` e2e test applies the policy and a real fixture through an actual Kyverno
installation on a `kind` cluster and asserts the live-mutated result, not just the static CLI
evaluation.

## Checklist

### DIFF
--- a/other/add-daemonset-tolerations/.chainsaw-test/chainsaw-test.yaml
+++ b/other/add-daemonset-tolerations/.chainsaw-test/chainsaw-test.yaml
@@ -1,0 +1,22 @@
+# yaml-language-server: $schema=https://raw.githubusercontent.com/kyverno/chainsaw/main/.schemas/json/test-chainsaw-v1alpha1.json
+apiVersion: chainsaw.kyverno.io/v1alpha1
+kind: Test
+metadata:
+  creationTimestamp: null
+  name: add-daemonset-tolerations
+spec:
+  steps:
+  - name: step-01
+    try:
+    - apply:
+        file: ../add-daemonset-tolerations.yaml
+    - assert:
+        file: policy-ready.yaml
+  - name: step-02
+    try:
+    - apply:
+        file: resource.yaml
+  - name: step-03
+    try:
+    - assert:
+        file: patched-resource.yaml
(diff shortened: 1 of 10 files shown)


yogeshbendre:
-Hi maintainers,

Just checking in on this one. It's been open about 2.5 weeks now and looks like the CI run is still waiting on approval to trigger for a first-time contributor. 
DCO is passing and it's still cleanly mergeable against main.

Happy to make any changes if there's feedback on the approach, or address anything that would help move it along — let me know. 
Thanks for your time. :)

PR #1528, opened 2026-08-24
fix: restrict require-probes vpol to regular and sidecar containers

p-fruck:
## Related Issue(s)

N/A

## Description

In https://github.com/kyverno/policies/pull/1432 more vpol examples have been added. One of them was `require-pod-probes`. Wheres the CPol only applies to regular containers, the VPol applies to ephemeral and initContainers as well. As described by the [docs](https://kubernetes.io/docs/concepts/workloads/pods/init-containers/#differences-from-regular-containers), init containers and ephemeral container don't support such probes, hence the policy will always fail if there is some initContainer specified.

This PR excludes ephemeral containers from the policy and only applies to initContainers which have `restart: always` set (sidecars), as they support probes. This implementation therefore also results in a different behavior compared to the old CPol.

## Checklist

### DIFF
--- a/best-practices-vpol/require-probes/require-probes.yaml
+++ b/best-practices-vpol/require-probes/require-probes.yaml
@@ -12,8 +12,10 @@ metadata:
       deployments, restarts, and upgrades. Liveness probes help the kubelet determine
       when to restart containers, while readiness probes help Services and Deployments
       determine when Pods are ready to receive traffic. Startup probes provide additional
-      control during container startup. This policy validates that all containers have
-      at least one of livenessProbe, readinessProbe, or startupProbe defined.      
+      control during container startup. This policy validates that all containers and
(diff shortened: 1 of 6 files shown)


best-practices-vpol/require-probes/require-probes.yaml
```diff
@@ -35,11 +37,22 @@ spec:
       resources: ["jobs", "cronjobs"]
       operations: ["CREATE", "UPDATE"]
   variables:
+  # Resolve the pod spec for both bare Pods and pod controllers.
+  - name: podSpec
+    expression: >-
+      has(object.spec.containers) ?
+        object.spec :
+        object.spec.template.spec
```
tom1299, line 45: Not sure whether this will work for cronjobs 🤔  Cronjobs will have pods at `spec.jobTemplate.spec.template.spec`. Or am I wrong here.

PR #1531, opened 2026-08-28
Add policy to block setting .host field from probes and lifecycle hooks

jayme-github:
## Description

In [0] the baseline pod security standard was extended to block setting the `.host` field in probes and lifecycle hooks.
This change adds the equivalent CEL policy.

[0] https://github.com/kubernetes/enhancements/issues/4940

## Checklist


jayme-github:
AFAICT all ci failures are for policies not related to this change

